PR #3, opened 2017-09-20
Start session on login

seromenho:
Starts a user session.


squallstar:
does it basically just call the session.get before the rest? just checking before looks like a lot of code changed but my guess is you just indented it.

seromenho:
@squallstar Yes, you are right. I should have mentioned it. It's just getting session + indentation.

seromenho:
https://github.com/Fliplet/fliplet-studio/issues/1443

PR #17, opened 2018-01-05
Use session to redirect on init

seromenho:
- On init check if user is logged in using the session instead of PV’s
- Remove app access validation

### DIFF
--- a/js/build.js
+++ b/js/build.js
@@ -164,9 +164,6 @@ $('[data-login-id]').each(function() {
           
           validateWeb()
             .then(function() {
-              return validateAppAccess();
-            })
-            .then(function() {
               if (Fliplet.Env.get('disableSecurity')) {
                 console.warn('Fliplet Login component tried to navigate to a page, but security is disabled.');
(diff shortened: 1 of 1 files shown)


js/build.js
```diff
@@ -159,61 +151,46 @@ $('[data-login-id]').each(function() {
...
-            if (Fliplet.Env.get('disableSecurity')) {
-              console.warn('Fliplet Login component tried to navigate to a page, but security is disabled.');
+          if (!Fliplet.Navigator.isOnline() && !Fliplet.Env.get("disableSecurity")) {
+            Fliplet.Navigate.to(_this.data.action);
+            return;
+          }
+          
+          validateWeb()
+            .then(function() {
+              return validateAppAccess();
```
tonytlwu: @seromenho Should you not be removing this too if you're removing the check for app access against the user?

seromenho: Totally right. Was already done. but with the loss of code and then recovering it I missed that. 
Thanks!! 
Done.

PR #30, opened 2018-03-20
Removed Fliplet.Security from being used. Not needed anymore.

Joao-pina-fernandes:
When the session work was done probably someone forgot to remove all the
Fliplet.Security logic


tonytlwu:
I don't know enough about this, either. Perhaps this was kept for backward compatibility.

tonytlwu:
Check and see if @seromenho has any opinion on this?

PR #37, opened 2018-07-17
refactor to use again the flipletLogin passport

squallstar:
Fixes an issue where the login component thinks users are already logged in when they're not.

The issue was recently discovered in the **NWP Admin Web App**.

---

ref https://github.com/Fliplet/fliplet-studio/issues/2847

### DIFF
--- a/js/build.js
+++ b/js/build.js
@@ -372,22 +372,37 @@ $('[data-login-id]').each(function() {
     return Promise.all(promises);
   }
 
+  function getUserFromSession(session) {
+    var loggedUser = session.server.flipletLogin;
+    if (Array.isArray(loggedUser)) {
+      loggedUser = loggedUser[0];
+    }
+
+    if (!loggedUser) {
+      throw new Error('Login session not found');
+    }
+
+    return loggedUser;
+  }
+
   function init() {
     _this.loginPV = {};
     Fliplet.User.getCachedSession()
       .then(function(session) {
-        if (!session || !session.user) {
+        if (!session || !session.server || !session.server.flipletLogin) {
           return Promise.reject('Login session not found');
         }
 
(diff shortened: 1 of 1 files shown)


js/build.js
```diff
@@ -372,22 +372,32 @@ $('[data-login-id]').each(function() {
...
-          id: session.user.id,
+          id: loggedUser.id || session.user.id,
           region: session.auth_token.substr(0, 2),
-          userRoleId: session.user.userRoleId,
+          userRoleId: loggedUser.userRoleId || session.user.userRoleId,
           authToken: session.auth_token,
-          email: session.user.email,
-          legacy: session.legacy
+          email: loggedUser.email || session.user.email,
+          legacy: loggedUser.legacy || session.legacy
```
tonytlwu: @squallstar Can you verify if the legacy property is non-empty when logging in with an account with a legacy v1 user? If it's empty then Fliplet Viewer will crash when opening v1 apps.

squallstar: @tonytlwu will do, is this for FV 2.0? If so, my guess is I need to rename `loggedUser.legacy` to `legacyId` and make it a boolean

tonytlwu: @squallstar `legacy` is expected to be an object

https://github.com/Fliplet/fliplet-ios-version-2/blob/f3b74606b74672206a91e19dadec8c79685b8b74/Fliplet/Classes/FLTAuthenticationManager.m#L50-L54

tonytlwu: It might require a server side change because that could be why I was using `session.user`. It's all coming back to me now 🥇

squallstar: ahhh thanks for the heads up, @tonytlwu yeah I think it will need a server-side change. I'll look into this.

PR #50, opened 2020-02-19
User will not lose components when change they position in the new D&D

YaroslavOvdii:
@sofiiakvasnevska 

## Issue
https://github.com/Fliplet/fliplet-studio/issues/5354

## Description
Changed components so they work with Fliplet.Widget.instance now.

## Screenshots/screencasts
https://share.getcloudapp.com/Qwu7RnDj

## Backward compatibility

This change is fully backward compatible.

### DIFF
--- a/css/build.css
+++ b/css/build.css
@@ -171,3 +171,8 @@ input.form-control:focus {
 		padding: 0 12.5%;
 	}
 }
+
+/* Fix component view when it is placed in the container */
+[data-widget-package="com.fliplet.login"] {
+  width: 100%;
+}
--- a/js/build.js
+++ b/js/build.js
@@ -1,4 +1,4 @@
-$('[data-login-id]').each(function() {
+Fliplet.Widget.instance('login', function(data) {
   var _this = this;
   var TWO_FACTOR_ERROR_CODE = 428;
   var ONE_TIME_2FA_OPTION = 'onetime';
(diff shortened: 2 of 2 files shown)


css/build.css
```diff
@@ -171,3 +171,8 @@ input.form-control:focus {
 		padding: 0 12.5%;
 	}
 }
+
+/* Fix component view when it is placed in the container */
```
squallstar: @tonytlwu @hcarneiro please confirm you're happy with this change

tonytlwu: Can't confirm if this would have any negative impact, but the code change looks legit. The rest is up to the developer and QA to verify.

js/build.js
```diff
@@ -12,8 +12,7 @@ $('[data-login-id]').each(function() {
     sendProcessing: 'Sending...'
   };
   _this.$container = $(this);
-  _this.id = _this.$container.attr('data-login-id');
```
tonytlwu: @YaroslavOvdii Can you confirm that `_this.id` is not used anywhere in this file?

css/build.css
tonytlwu: @hcarneiro Is this necessary because D&D3 now has all content laid out like inline blocks using flexbox, and therefore this block DIV needs `width: 100%` to occupy the 100% of horizontal space?

js/build.js
YaroslavOvdii: @tonytlwu Yes, `_this.id` do not use anywhere in this file
It was used only to get a `_this.data` from `Fliplet.Widget.getData(_this.id)`

css/build.css
hcarneiro: Yes, this will force the component to occupy the entire horizontal space, by default on the new D&D system. This can be overwritten by the Appearance Settings.
I'm fine with this change!

tonytlwu: @hcarneiro Thanks!

PR #57, opened 2021-11-11
ID-1349 [Maintenance] Linting and cleanup

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-1349

Preparation before adding validation of user sessions

### DIFF
--- a/js/build.js
+++ b/js/build.js
@@ -593,7 +593,7 @@ Fliplet.Widget.instance('login', function(data) {
       })
       .then(function() {
         if (!Fliplet.Navigator.isOnline()) {
-          return Promise.resolve();
+          return;
         }
 
         return validateWeb()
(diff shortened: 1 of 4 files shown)


js/build.js
```diff
@@ -10,17 +10,13 @@ Fliplet.Widget.instance('login', function(data) {
...
     sendProcessing: 'Sending...',
-    continuDefault :'Continue',
+    continueDefault: 'Continue',
     continueProcessing: 'Please wait...'
   };
+
   _this.$container = $(this);
   _this.data = data;
   _this.pvNameStorage = 'fliplet_login_component';
-  var studioUrls = {
```
squallstar, line 19: I'm guessing these aren't required anymore?

squallstar (approved):

js/build.js
tonytlwu: That's right. I don't see any reference of it.

PR #59, opened 2021-11-17
ID-1349 [Feat] Validate accounts and T&Cs on all screens

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-1349

Also refactors build.js to use `Fliplet.Login` functions defined in `app.js`.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -219,7 +219,7 @@
     expire: 60 * 60 * 12 // Keep cache for half a day
   }, function onFetchData() {
     return Fliplet.Login.validateAccount().catch(function() {
-      // If it needs to be completed but not properly completed, clear cache
+      // Validation was not completed, clear cache to make sure the check continues
       Fliplet.Cache.remove(cacheKey);
     });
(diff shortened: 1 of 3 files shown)


js/app.js
```diff
@@ -0,0 +1,246 @@
...
+     * @param {String} data.url - Request URL
+     * @param {String} data.token - Request authentication token
+     * @returns {Promise} Promise is resolved when the request completes
+     */
+    function request(data) {
+      data = data || {};
+
+      // Validate token
+      return Fliplet.Navigator.onReady().then(function() {
+        data.url = Fliplet.Env.get('apiUrl') + data.url;
```
squallstar, line 76: you should be able to use

```js
Fliplet.API.request({
  url: data.url,
  headers: {
    'Auth-token': data-token
  }
})
```

tonytlwu: @squallstar Thanks. Updated.

squallstar (approved):

PR #75, opened 2023-12-13
ID-3687 [Fix] Check for valid credential types before determining if the user is linked to SSO

tonytlwu:
Ref. https://weboo.atlassian.net/browse/ID-3687

Skip compile: 
Includes translation: 
Initiative: Fliplet login issue
Impact: High
Affected: Fliplet Viewer users
Verify: Users can login and remain logged in

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -117,7 +117,10 @@
 
         data = data || {};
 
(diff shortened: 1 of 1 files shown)


simonchapman1986 (approved):
Risk: medium
Affected: SSO
Verify: 
- users can login correctly for apps with SSO
- users have valid credential types

PR #77, opened 2024-04-29
ID-4115 [FIX] Fliplet Login form disappears from the screen when hidden DOM element gets focus

armine-fliplet:


dwfliplet (approved):

simonchapman1986 (approved):
Risk: low
Affected: login form
Verify: login form doesn't disappear from screen

PR #81, opened 2024-07-25
ID-4114 [FIX] Add check to ensure initSession won’t navigate back to different screen when user trys to logout

armine-fliplet:
### Result 
[Fliplet Login.webm](https://github.com/user-attachments/assets/58a9f73f-f233-4b5c-a2a6-aeebf6d7e02f)

### DIFF
--- a/js/build.js
+++ b/js/build.js
@@ -516,8 +516,10 @@ Fliplet.Widget.instance('login', function(data) {
     }
 
     function initSession() {
+      let preserveSession;
       Fliplet.User.getCachedSession()
         .then(function(session) {
+          preserveSession = session;
           var passport = session && session.accounts && session.accounts.flipletLogin;
 
           if (passport) {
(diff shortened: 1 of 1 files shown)


js/build.js
```diff
@@ -555,11 +555,17 @@ Fliplet.Widget.instance('login', function(data) {
             return Promise.reject(T('widgets.login.fliplet.warnings.noRedirectWhenEditing'));
           }
 
-          var navigate = Fliplet.Navigate.to(_this.data.action);
+          return Fliplet.Session.get().then(function(session) {
+            if (session.client && session.client.source === 'studio') {
+              return Promise.reject('Logout prevented in preview mode.');
```
zeryabkhan91, line 560: @armine-fliplet This reject error is confusing in a sense because when a user is logging out in Preview mode, we are destroying session passport from backend which is equivalent to logging out. Error message should be something like: 'Preventing navigation to another screen in Preview mode.'

zeryabkhan91 (dismissed):

Arpanexe (approved):
LGTM

dwfliplet (approved):

PR #89, opened 2026-02-26
feat(DEV-935): Replace auth_token URL param with one-time state token in auth redirect

zeryabkhan91:
## Summary

- Exchanges the real session token for a short-lived state token (AES-encrypted, 60s expiry) before opening the auth redirect URL in the in-app browser
- The `auth_token` no longer appears in the IAB URL, removing exposure via browser history, proxy logs, and Referrer headers

## Test plan

- [ ] Trigger the Fliplet Login auth redirect flow (e.g. account verification in-app browser)
- [ ] Inspect the URL opened in the in-app browser — should contain `?state=<hex_string>` and **no** `auth_token`
- [ ] Complete the full redirect flow and confirm the user is authenticated/verified successfully
- [ ] Verify `POST /v1/session/authorize/state` is called and returns a state token before navigation

## Dependencies

Requires `fliplet-api` changes from DEV-935 (`POST /v1/session/authorize/state` endpoint and state token resolution in `libs/authenticate.js`).

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -163,6 +163,20 @@
               Fliplet.Navigate.defaults.disableShare = defaultShare;
             });
           });
+        }).catch(function(err) {
+          console.error('[Fliplet.Login] Failed to obtain state token', err);
+          Fliplet.Navigate.defaults.disableShare = defaultShare;
+          return new Promise(function(resolve, reject) {
+            Fliplet.Navigate.url({
+              url: (Fliplet.Env.get('primaryApiUrl') || Fliplet.Env.get('apiUrl')) + 'v1/auth/redirect?auth_token=' + storage.auth_token + '&utm_source=com.fliplet.login',
+              inAppBrowser: true,
+              onclose: function() {
+                validateAccount().then(resolve).catch(reject);
+              }
+            }).then(function() {
+              Fliplet.Navigate.defaults.disableShare = defaultShare;
+            });
+          });
         });
       });
     }
--- a/js/app.js
+++ b/js/app.js
@@ -146,15 +146,22 @@
 
         Fliplet.Navigate.defaults.disableShare = true;
 
-        return new Promise(function(resolve, reject) {
-          Fliplet.Navigate.url({
-            url: (Fliplet.Env.get('primaryApiUrl') || Fliplet.Env.get('apiUrl')) + 'v1/auth/redirect?auth_token=' + storage.auth_token + '&utm_source=com.fliplet.login',
-            inAppBrowser: true,
-            onclose: function() {
-              validateAccount().then(resolve).catch(reject);
-            }
-          }).then(function() {
-            Fliplet.Navigate.defaults.disableShare = defaultShare;
+        // Exchange the real session token for a one-time state token so that
+        // auth_token never appears in the URL opened in the in-app browser.
+        return Fliplet.API.request({
+          url: 'v1/session/authorize/state',
+          method: 'POST'
+        }).then(function(response) {
+          return new Promise(function(resolve, reject) {
+            Fliplet.Navigate.url({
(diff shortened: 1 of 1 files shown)


Komsomol (changes requested):
## DEV-935 Login Widget Review — Request Changes (minor)

Clean, well-scoped change. The security improvement is correct — removing `auth_token` from the in-app browser URL eliminates exposure via browser history, Referrer headers, and proxy logs.

### One issue to fix

**Missing `.catch()` handler on state token request**

```js
return Fliplet.API.request({
  url: 'v1/session/authorize/state',
  method: 'POST'
}).then(function(response) { ... });
```

If the API call fails (network error, Redis down, endpoint not yet deployed), the promise rejects with no handler. The `defaultShare` flag set on line 148 is never restored, and the user gets no feedback about why account setup failed.

Add a `.catch()` that:
1. Restores `Fliplet.Navigate.defaults.disableShare = defaultShare`
2. Falls back to the legacy `auth_token` URL approach for graceful degradation during rollout

```js
}).catch(function(err) {
  console.error('[Fliplet.Login] Failed to obtain state token', err);
  Fliplet.Navigate.defaults.disableShare = defaultShare;
  // Fallback: use auth_token directly
  return new Promise(function(resolve, reject) {
    Fliplet.Navigate.url({
      url: (Fliplet.Env.get('primaryApiUrl') || Fliplet.Env.get('apiUrl'))
        + 'v1/auth/redirect?auth_token=' + storage.auth_token
        + '&utm_source=com.fliplet.login',
      inAppBrowser: true,
      onclose: function() {
        validateAccount().then(resolve).catch(reject);
      }
    }).then(function() {
      Fliplet.Navigate.defaults.disableShare = defaultShare;
    });
  });
});
```

This fallback can be removed once the `auth_token` query param is fully deprecated.

### Deployment note

This widget must deploy AFTER API PR #7683 (which provides the `POST /v1/session/authorize/state` endpoint). If the widget ships first, the API call will 404 and without the `.catch()` above, account setup will break silently.

Komsomol (approved):
**APPROVE** — Clean implementation with proper fallback.

Minor suggestion: the success and catch paths duplicate the `Fliplet.Navigate.url()` block (~15 lines each), differing only by `state=` vs `auth_token=`. Consider extracting a small helper to reduce drift risk, but not blocking.

farhantariq12b (approved):

Komsomol:
## Post-Merge Review — DEV-935 Login Widget

### MEDIUM — `disableShare` Restored Too Early in Catch Path

In the `.catch()` branch, `Fliplet.Navigate.defaults.disableShare = defaultShare` runs **before** the IAB opens, then runs again in the `.then()` after `Navigate.url()`. The early restoration defeats the purpose of disabling share during the IAB flow. Fix: remove the early `disableShare` restoration and rely only on the one inside the `.then()` callback.

### LOW — Code Duplication

The success and catch paths duplicate ~15 lines of `Fliplet.Navigate.url()` logic, differing only by `state=` vs `auth_token=`. The SAML2 widget (PR #42) handles this more cleanly with a ternary `authParam` variable and a single navigation block. Consider refactoring to match that pattern:

```javascript
}).then(function(response) {
  var authParam = response.state
    ? 'state=' + response.state
    : 'auth_token=' + storage.auth_token;

  return new Promise(function(resolve, reject) {
    Fliplet.Navigate.url({
      url: baseUrl + 'v1/auth/redirect?' + authParam + '&utm_source=com.fliplet.login',
      // ...
    });
  });
});
```

### LOW — No Validation on `response.state`

No check that `response.state` is truthy before concatenating into the URL. If the API returns `{ state: undefined }`, the URL becomes `?state=undefined`.

---

The security improvement is effective — `auth_token` replaced with a short-lived AES state token in the IAB URL. Fallback to legacy behaviour on API failure is the right call for reliability.

PR #91, opened 2026-04-09
DEV-1090: [Refactor] Replace inline auth with popup-based unified flow

hcarneiro:
## Summary

Replaces the widget's 7-state inline auth form with a single **Sign in** button that delegates to the unified `/v1/auth/login` page (rendered by the fliplet-api auth-loader script).

**Sign-in dispatch by context:**
- **Deployed web app** (`platform === 'web'`, not Studio interact) — top-level redirect to the auth-loader; the auth-loader redirects back to the same page with `token` + `user` in the query string, which the widget strips and processes on mount.
- **Studio interact mode** — popup (defensive fallback; the Sign in button is normally disabled in interact anyway).
- **Cordova native app** — InAppBrowser with a sentinel callback URL; same contract the CLI / VSCode extension use.

The widget's existing post-login glue (`Fliplet.Login.updateUserStorage`, `Fliplet.Hooks.run('login', ...)`, `Fliplet.Login.validateAccount`, `Fliplet.Navigate.to(_this.data.action)`) runs unchanged regardless of which dispatch ran.

This unifies the sign-in experience across Studio, the API protected-route page, the Fliplet Login widget, and the future CLI consumer — auth UX changes only need to be made in one place going forward.

Pairs with:
- Fliplet/fliplet-api branch `v3/DEV-908-signup-sso` (the unified `/v1/auth/login` page changes + new `hideModeSwitch` / `hideRememberMe` auth-loader options + per-app origin allow-list)

Spec doc: `docs/superpowers/specs/2026-04-09-fliplet-login-widget-redesign.md` in the fliplet-api repo.

## Changes

**Form rewrite:**
- `build.html`: 343 lines → 29 lines. Single Sign in button + heading + loader + offline notification + error holder.
- `js/build.js`: adds `openSignInSameTab` (deployed web), `openSignInPopup` (Studio interact fallback), `openSignInIAB` (Cordova native), and `handleAuthSuccess` convergence point. Adds `handleSameTabReturn` to catch the return-leg query params on mount and strip them via `history.replaceState`. Preserves `initSession` (already-signed-in detection), `Fliplet.Login.*` glue, the login hook, the `LOGIN_FLAG_KEY` storage flag, Studio preview reload handlers, and offline detection.
- `js/build.js`: uses `Fliplet.Env.get('primaryApiUrl')` first, falling back to `apiUrl`, so the redirect/popup opens against the canonical API host rather than the apps-host-proxied URL (which doesn't serve `/v1/auth`).
- `css/build.css`: trimmed inline-form rules, added `.fliplet-login-button` styles. 203 lines → ~140 lines.

**Heading default:**
- Restored the `{{else}}` fallback in `build.html` so the default "Welcome to the login of this app" renders on first widget add. Regression fix for a removed fallback that left the widget with no heading until the user opened the Interface panel.

**Settings panel:**
- Adds an optional **Sign-in button label** text input (defaults to "Sign in"). Existing **Login Heading** + **Redirect after logging in** settings unchanged.

**Translations:**
- New keys: `actions.signIn`, `actions.waitingForSignIn`, `errors.popupBlocked`. Need to be translated for non-English locales via the standard Lokalise pipeline.

## Backwards compatibility

| Surface | Status |
| --- | --- |
| `data.heading` setting | ✅ Unchanged |
| `data.action` (next-page navigation) | ✅ Unchanged |
| `data.buttonLabel` setting | ✅ New, optional |
| `Fliplet.Hooks.on('login', cb)` | ✅ Fires after the new flow with the same payload shape |
| `Fliplet.Login.updateUserStorage` | ✅ Called with the same fields |
| `Fliplet.Login.validateAccount` → `goToAccountSetup` | ✅ Unchanged — `app.js` is untouched |
| `Fliplet.Profile.set`, `Fliplet.Navigate.to` | ✅ Unchanged |
| Already-signed-in detection (`initSession`) | ✅ Unchanged — skips the button when a session is cached |
| LOGIN_FLAG_KEY storage flag | ✅ Set after successful login (App List component still works) |

**Breaking-ish:** customers who applied custom CSS targeting the inline form selectors (`.login-form`, `.login_email`, `.btn-fliplet-login`, etc.) will see those elements no longer exist. Custom code listening for inline form events should switch to the `Fliplet.Hooks.on('login', cb)` hook.

### Desktop (deployed web app)
- [ ] Load app on `*.fliplet.com` → click Sign in → **browser navigates to `/v1/auth/login`** (no popup) → complete sign-in → browser redirects back to the original page → `token` / `user` are stripped from the URL via `history.replaceState` → app navigates to the configured next page
- [ ] Same flow with Google SSO
- [ ] Same flow with SAML2 (if a test IdP is configured)
- [ ] Custom-domain app: load on a verified custom domain → all flows work
- [ ] Unverified custom-domain app: load → click Sign in → auth-loader shows the origin error block
- [ ] Already signed in: load app while session is cached → button skipped, navigates straight to next page
- [ ] Repeat sign-in on the same page (sign in, log out, sign in again) → callback URL doesn't accumulate stale `token` / `user` params
- [ ] Fresh widget add in Studio → drop on a new screen → heading "Welcome to the login of this app" renders immediately, without opening the Interface panel

### Mobile native (Cordova)
- [ ] iOS + Android: tap Sign in → IAB opens → email/password sign in → IAB closes → app navigates
- [ ] iOS + Android: tap Sign in → IAB opens → SAML2 IdP round-trip → IAB closes → app navigates
- [ ] Already signed in on launch: button skipped
- [ ] User closes IAB mid-flow: button stays active, no error toast

### Studio preview / interact
- [ ] Drag widget into a screen → settings panel renders correctly with heading, action, and new button label inputs
- [ ] Save settings → refresh preview → widget shows updated heading and button label
- [ ] Sign in button is disabled in Studio interact mode (existing behavior)
- [ ] If interact-mode disable is bypassed, click falls back to popup (not same-tab) so the Studio iframe isn't hijacked


farhantariq12b (changes requested):
## Code Review — Requesting Changes

Major architectural shift from inline forms to popup-based auth — the simplification is welcome and the unified flow is the right direction. A few security and robustness issues to address:

### 1. HIGH — postMessage must also validate `event.source`

`js/build.js` ~line 925-927:

```js
function onMessage(event) {
  if (event.origin !== apiOrigin) return;
```

This checks origin but not that the message came from *the popup we opened*. Any page hosted on the API origin could send a matching `fliplet-login-success` message. The `popup` variable is in scope — add:

```js
if (event.source !== popup) return;
```

### 2. HIGH — No timeout on popup auth flow

If the popup loads but the user abandons it without closing (switches tabs, loses connectivity mid-auth), the widget hangs in loading state indefinitely. The only recovery is the 500ms poll detecting `popup.closed`, which won't fire if the popup is still open.

Add a timeout (e.g. 10 minutes) that calls `cleanup()`, closes the popup, and shows a "Sign-in timed out" message.

### 3. HIGH — Concurrent popup re-entry not prevented

If the user double-clicks the sign-in button before the first popup opens, two popups open with two `onMessage` listeners registered. Both will process the same postMessage, causing `handleAuthSuccess` to run twice (double navigation, double storage writes).

**Fix:** Either disable the button immediately on first click (before the async popup open), or track an `isAuthInProgress` flag and bail early if already set, or close any existing popup before opening a new one.

### 4. MEDIUM — `getApiOrigin()` fallback doesn't return a valid origin

Lines 765-770: The `catch` branch does `getApiHost().replace(/\/$/, '')` which strips trailing slash but doesn't extract `scheme+host+port`. If `getApiHost()` returns a URL with a path (e.g. `https://api.fliplet.com/v1`), the origin check at line 927 will compare against the full URL, never matching any `event.origin`.

**Fix:** Parse explicitly in the fallback, or ensure `getApiHost()` always returns a root URL.

### 5. MEDIUM — IAB double-callback not guarded

In `openSignInIAB()`, the `loadstart` handler fires on every navigation. If the IAB hits the callback URL, then the user taps back and it navigates there again, `handleAuthSuccess` fires twice. The IAB is closed at line 1044, but there's a race if two `loadstart` events queue before the close takes effect.

**Fix:** Add a `handled` flag set to `true` on first callback match.

### 6. LOW — Breaking CSS selectors should be documented

The PR body mentions "breaking CSS selectors" but doesn't list them. Customers with custom CSS targeting `.login-form`, `.btn-fliplet-login`, `.state[data-state=*]`, `input.login_email`, `.fl-two-auth` etc. will see styles silently break. Please add a migration note listing removed selectors and their replacements (`.fliplet-login-button`, etc.).

---

The overall direction is solid — unifying auth into one place is the right call. The main blockers are the `event.source` check (#1) and re-entry prevention (#3).

build.html
```diff
@@ -2,342 +2,26 @@
...
-      </h3>
-      <p data-translate="widgets.login.fliplet.update.instruction">
-        Your organization has updated its password policy and you will need to update your password
-        to keep meeting the security requirements.
+    <button
+      type="button"
+      class="btn btn-primary fliplet-login-button"
+      data-translate="widgets.login.fliplet.actions.signIn"
+    >
+      {{#if buttonLabel}}{{buttonLabel}}{{else}}Sign in{{/if}}
```
tonytlwu, line 19: Fallback should use translation

farhantariq12b (approved):

PR #92, opened 2026-04-20
DEV-983 [Feat] Restrict Dev Environment Dashboard access to Fliplet Admins

zeryabkhan91:
## Summary
- Adds `Fliplet.Login.verifyUserForDevEnvApp(user)` in `js/app.js` — logs the user out and rejects with an i18n error when the current app is a known Dev Environment Dashboard but the user's `userRoleId !== 1` (Fliplet Admin). Resolves silently otherwise.
- Invokes the check from both login paths in `js/build.js` (SSO `onclose` and the password-based `login()` flow) so authentication actually halts on failure (earlier version called an undefined `reject` and didn't `return` the promise).
- Invokes the check inside `validateAccount` in `js/app.js` so returning sessions are also enforced on app boot (validateAccount runs via `Fliplet.Cache.get` every 12h), not only at login time.
- Adds translation key `widgets.login.fliplet.errors.userNotAFlipletAdmin`.

## Why
The Dev Environment Dashboard web app uses this shared Fliplet login widget. The `/v1/dev-envs` API is already admin-gated server-side (`router.use(devEnvironment.authenticateAdminWithFlipletLogin)`), but without a client-side check a non-admin could still complete login and land in a broken dashboard UI. This closes the UX gap.

## Notes / follow-ups
- `DEV_ENVIRONMENT_APPS` is hardcoded (master + production app IDs per environment URL). Adding a future dev-env deployment requires updating this map and cutting a widget release. A widget-data-driven `requireFlipletAdmin` flag would be cleaner — happy to follow up in a separate ticket.
- Please double-check the `https://apps.fliplet.test/` host is the correct URL for that environment.

## Test plan
- [ ] Log in as a Fliplet Admin (`userRoleId === 1`) on the Dev Env Dashboard app — access granted.
- [ ] Log in as a non-admin on the Dev Env Dashboard app — toast shows "Only Fliplet Admins can access this app.", session is logged out.
- [ ] Repeat for the SSO login path.
- [ ] Log in as a non-admin on a non-dev-env app — unaffected, logs in normally.
- [ ] Returning session: non-admin with an existing session reopens the Dev Env Dashboard app — blocked on boot via `validateAccount`.
- [ ] Translation renders correctly.

### DIFF
--- a/js/app.js
+++ b/js/app.js
@@ -3,8 +3,17 @@
 
   Fliplet.Login = (function() {
     var ORG_ADMIN_ROLE_ID = 1;
+    var FLIPLET_ADMIN_ROLE_ID = 1;
     var storageName = 'fliplet_login_component';
     var skipSetupStorageName = 'skipFlipletAccountSetup';
+
+    // Master app and production app IDs for the deployed Dev Environment Dashboards.
+    // Kept in sync with the map in js/build.js — update both when adding a new env.
+    var DEV_ENVIRONMENT_APPS = {
+      'https://env.fliplet.com/': [436446, 436447],
+      'https://staging-apps.fliplet.com/': [511849, 511850],
+      'https://apps.fliplet.test/': [11, 12]
+    };
 
     /**
      * Creates user profile data
@@ -90,6 +99,28 @@
 
           return response;
         });
+      });
+    }
+
+    /**
+     * Verify that the given user is allowed to access the current app when it
+     * is a Dev Environment Dashboard (Fliplet Admins only). Logs the user out
+     * and rejects on failure; resolves silently otherwise.
+     * @param {Object} user - User object exposing userRoleId
+     * @returns {Promise} Resolves if allowed, rejects with i18n error if not
+     */
+    function verifyUserForDevEnvApp(user) {
+      var allowedAppIds = DEV_ENVIRONMENT_APPS[Fliplet.Env.get('appsUrl')];
+      var isDevEnvApp = Array.isArray(allowedAppIds)
+        && allowedAppIds.indexOf(Fliplet.Env.get('appId')) !== -1;
+
+      if (!isDevEnvApp || !user || user.userRoleId === FLIPLET_ADMIN_ROLE_ID) {
+        return Promise.resolve();
+      }
+
+      // Tear down the session so the user isn't left authenticated-but-blocked
+      return Fliplet.Session.logout().catch(function() {}).then(function() {
+        return Promise.reject(T('widgets.login.fliplet.errors.userNotAFlipletAdmin'));
       });
     }
 
@@ -195,6 +226,10 @@
       }
 
       return getData.then(function(response) {
+        return verifyUserForDevEnvApp(_.get(response, 'user')).then(function() {
+          return response;
+        });
+      }).then(function(response) {
         return userMustSetupAccount(response).then(function(setupRequired) {
           if (setupRequired) {
             return goToAccountSetup();
@@ -217,6 +252,7 @@
     return {
       updateUserStorage: updateUserStorage,
       validateAccount: validateAccount,
+      verifyUserForDevEnvApp: verifyUserForDevEnvApp,
       setSkipSetupStorage: setSkipSetupStorage
     };
   }());
--- a/js/app.js
+++ b/js/app.js
@@ -8,7 +8,6 @@
     var skipSetupStorageName = 'skipFlipletAccountSetup';
 
     // Master app and production app IDs for the deployed Dev Environment Dashboards.
-    // Kept in sync with the map in js/build.js — update both when adding a new env.
     var DEV_ENVIRONMENT_APPS = {
       'https://env.fliplet.com/': [436446, 436447],
       'https://staging-apps.fliplet.com/': [511849, 511850],
--- a/js/app.js
+++ b/js/app.js
@@ -2,8 +2,8 @@
   var Fliplet = window.Fliplet || {};
 
   Fliplet.Login = (function() {
-    var ORG_ADMIN_ROLE_ID = 1;
-    var FLIPLET_ADMIN_ROLE_ID = 1;
+    var ORG_ADMIN_ROLE_ID = 1; // This role ID is assigned to Fliplet Studio user who is an organization admin
+    var FLIPLET_ADMIN_ROLE_ID = 1; // This is user role ID which is only assigned to Fliplet Staff
     var storageName = 'fliplet_login_component';
     var skipSetupStorageName = 'skipFlipletAccountSetup';
 
@@ -121,7 +121,11 @@
       return Fliplet.Session.logout().catch(function(err) {
         console.warn('[verifyUserForDevEnvApp] Failed to log out blocked user:', err);
       }).then(function() {
-        return Promise.reject(T('widgets.login.fliplet.errors.userNotAFlipletAdmin'));
+        var error = new Error(T('widgets.login.fliplet.errors.userNotAFlipletAdmin'));
+
+        error.code = 'USER_NOT_FLIPLET_ADMIN';
+
+        return Promise.reject(error);
       });
     }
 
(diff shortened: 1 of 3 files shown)


hcarneiro (changes requested):
## Summary
Adds a client-side Fliplet-Admin gate on the shared login widget when the current app is a Dev Environment Dashboard. Introduces `Fliplet.Login.verifyUserForDevEnvApp(user)` in `js/app.js`, wires it into the SSO `onclose` path, the password-based `login()` path, and `validateAccount()` (so returning sessions are re-checked on boot), and adds a `userNotAFlipletAdmin` translation. Complements the existing server-side admin gate on `/v1/dev-envs` so non-admins can't land in a broken UI.

## Critical Issues (Must Fix Before Merge)

1. **SSO path passes the wrong object to the admin check** — `js/build.js` (SSO `onclose` handler).
   The SSO branch calls `Fliplet.Login.verifyUserForDevEnvApp(user)` where `user = _.get(session, 'server.passports.flipletLogin', [])[0]`. In the very same block, the subsequent `updateUserStorage` reads `userRoleId: session.user.userRoleId` — not `user.userRoleId` — which is the established pattern in this file because the passport object doesn't reliably carry `userRoleId`. If the passport object lacks `userRoleId`, `user.userRoleId === FLIPLET_ADMIN_ROLE_ID` evaluates to `undefined === 1` and **every SSO login on a dev-env app gets rejected, including real Fliplet Admins**. Pass `session.user` (or an object whose `userRoleId` is taken from `session.user`) instead, and cover this path explicitly in the QA test plan.

2. **Misleading sync comment** — `js/app.js` near the `DEV_ENVIRONMENT_APPS` declaration.
   `// Kept in sync with the map in js/build.js — update both when adding a new env` — there is no `DEV_ENVIRONMENT_APPS` map in `js/build.js`; the map only exists in `app.js`. Remove the comment (or actually split/share the list). As written it will mislead whoever next adds an environment.

## Warnings (Should Address)

1. **Placeholder-looking local test IDs** — `'https://apps.fliplet.test/': [11, 12]`. This looks like scaffolding; the PR description already flags it for verification. Confirm the URL + app IDs are correct (or drop the entry) before merging, since a wrong URL silently turns the gate off for that environment (`indexOf === -1`, check becomes a no-op) rather than failing loudly.
2. **`ORG_ADMIN_ROLE_ID` and `FLIPLET_ADMIN_ROLE_ID` are both `1`**. If they truly refer to the same role ID (platform-level admin), consolidate to one constant; if they're semantically distinct roles that happen to coincide today, add a one-line comment explaining why they're kept separate.
3. **Hardcoded app ID map requires a widget release per env** — already acknowledged in the PR description as a follow-up. A `widgetData`-driven `requireFlipletAdmin` flag (or a server-side `app.requiresFlipletAdmin` exposed via `Fliplet.Env`) would be cleaner and decouple widget releases from environment provisioning.
4. **No tests added** — no Jest coverage for the new gate, the three call sites, or the no-op-on-non-dev-env case. Given this is auth-adjacent and touches every login path, even one spec per branch would meaningfully reduce regression risk.

## Suggestions (Nice to Have)
- The `Fliplet.Session.logout().catch(function() {}).then(...)` sequence silently swallows logout failures; consider `console.warn`-ing the error so a stuck "authenticated-but-blocked" state is diagnosable.
- `isDevEnvApp` could early-return before touching the user object for readability: `if (!allowedAppIds || allowedAppIds.indexOf(Fliplet.Env.get('appId')) === -1) return Promise.resolve();`.
- Consider logging an analytics event on the rejection path — useful for detecting accidental lockouts after a role change or a bad app-ID map update.
- The Codacy check is currently in `action_required`; resolve or acknowledge before merge.

## Verdict
**Request changes** — the SSO path almost certainly passes a user object without `userRoleId` to the gate, which would lock real admins out via SSO. Fixing that plus the stale "kept in sync" comment is cheap; the rest can follow or be addressed in-line at the author's discretion.

hcarneiro (commented):
## Review summary

Adds a client-side guard that logs out and rejects non-admin users when they attempt to authenticate into any hardcoded Dev Environment Dashboard app (`env.fliplet.com`, `staging-apps.fliplet.com`, `apps.fliplet.test`). The check is wired into three entry points — SSO `onclose`, the password-based `login()` flow, and `validateAccount` (so returning sessions hit it on boot via the 12h `Fliplet.Cache.get`). Complements the server-side `authenticateAdminWithFlipletLogin` middleware on `/v1/dev-envs` as defence-in-depth for UX.

### Critical issues
None blocking. The guard works as specified and the three entry points are wired consistently.

### Warnings (should address)

1. **`ORG_ADMIN_ROLE_ID` and `FLIPLET_ADMIN_ROLE_ID` both `= 1` with no explanation** — `js/app.js:5-6`. These are different concepts (`organizationUser.organizationRoleId` vs `user.userRoleId`) that coincidentally share the value. The current naming suggests they're the same thing. At minimum add a one-line comment on `FLIPLET_ADMIN_ROLE_ID` clarifying it's the system-level role checked against `user.userRoleId` — otherwise the next reader will try to consolidate them.

2. **Silent failure on app boot if `validateAccount` rejects** — `js/app.js:224-231`. When `Fliplet.Cache.get` runs `Fliplet.Login.validateAccount()` on boot and `verifyUserForDevEnvApp` rejects, I don't see any `.catch` that surfaces the `userNotAFlipletAdmin` toast. The returning-session flow from the test plan will log the user out correctly, but they'll probably just see a blank/login screen with no explanation of *why*. Worth either surfacing a toast on that reject path or confirming the app's existing error plumbing handles it — the login-time paths have explicit reject handlers that lead to a UI toast; the boot path doesn't.

3. **Logout failure is swallowed** — `js/app.js:120-122`. If `Fliplet.Session.logout()` rejects we `console.warn` and still reject with the i18n message, leaving the user authenticated client-side. The server-side middleware is the real backstop, but consider whether to also clear `Fliplet.Profile` / app storage manually so the blocked user isn't left with a partial authenticated state that other widgets might read.

4. **Inconsistent user shape across call sites** — SSO passes `session.user` (extended with `passport[0]`); `login()` passes `user` which is `_.get(response, 'session.server.passports.flipletLogin', [])[0]`; `validateAccount` passes `_.get(response, 'user')` from `getUserData()`. The guard only reads `.userRoleId`, so as long as all three shapes expose it this works — but worth confirming that the raw passport entry in `login()` always includes `userRoleId` (commit c36d14a1 specifically addressed this for SSO, so likely fine, but worth double-checking the non-SSO path).

5. **No test coverage for the new gate** — security-adjacent logic with five branches (non-dev-env app → allow, missing user → allow, admin → allow, non-admin on dev-env app → reject, logout failure → still reject). A minimum of one happy and one sad path test with `Fliplet.Env` / `Fliplet.Session` mocks would be cheap insurance.

### Suggestions (nice to have)

- **Widget-data-driven flag** — the PR description already flags this as a follow-up. A `requireFlipletAdmin` boolean in the widget data schema would eliminate the hardcoded `DEV_ENVIRONMENT_APPS` map and the release-on-deploy-new-env coupling. Fine to defer as suggested.
- **Confirm `apps.fliplet.test` is the actual local dev host** — PR description explicitly asks for verification. A mismatch would silently disable the gate locally.
- **Missing-user safety**: `!user` currently short-circuits to `Promise.resolve()`. For a dev-env app with a missing user object you'd probably want to fail closed. Low priority since upstream callers validate `user` first, but defensible hardening.
- **Widen the error-toast regex** — unrelated to this PR, but `/cancelled|closed/i` in `build.js:194` will swallow any future error message containing those words. Not an issue today but worth being aware of.

### Verdict
Commenting, not blocking — core logic is correct, wiring covers all three entry points, and the `.catch(reject)` fix to the SSO chain is a nice side-cleanup. I'd like warnings #1 and #2 addressed and at least a minimal test for the new function before merge; the hardcoded apps map and missing-user short-circuit can ship as-is with follow-up tickets.

hcarneiro (approved):
Thanks for the quick turnaround — all three warnings addressed cleanly, and a couple of them better than I suggested:

- **#1** — clear comments on both `ORG_ADMIN_ROLE_ID` / `FLIPLET_ADMIN_ROLE_ID` distinguishing the two concepts. 👍
- **#2** — using a typed `Error` with `error.code = 'USER_NOT_FLIPLET_ADMIN'` and a dedicated `.catch` on `Fliplet.Cache.get` that toasts only on that code. Nicer than magic-string matching.
- **#3** — belt-and-suspenders clear of `Fliplet.App.Storage` and `Fliplet.Profile` after the logout attempt, each with their own `.catch` warn. Exactly the "no partial authenticated state" hardening I was after.
- **#4** — accepting the verbal confirmation that `userRoleId` is always present in the session passport.

Follow-up ticket for the widget-data-driven `requireFlipletAdmin` flag (to kill the hardcoded `DEV_ENVIRONMENT_APPS` map) and a minimal unit test for `verifyUserForDevEnvApp` would both be worthwhile at some point, but neither is a blocker here.

Approving. 🚀

galisufyan-327 (approved):

farhantariq12b (approved):

PR #98, opened 2026-05-27
release/DEV-1090

hcarneiro:
## Summary

Replaces the in-widget Fliplet email/password sign-in (with inline 2FA, forgot-password, password reset, and forced-password-update flows) with a single **Sign in** button that delegates to the unified auth-loader hosted on the Fliplet API (`/v1/auth/login`). All three sign-in paths converge on a shared `handleAuthSuccess({ token, user })`:

- **Web same-tab** — `window.location.assign('<api>/v1/auth/login?return=callback&callback=<page-url>&state=<nonce>')`. The auth-loader does a top-level redirect back to the page with `?token=...&user=...&state=...`. `handleSameTabReturn` validates the state nonce against `sessionStorage`, validates the user payload shape, and stamps the token into the local session.
- **Web popup** — Delegates to `Fliplet.Auth.signIn()` (SDK call) which opens the unified sign-in page in a popup and handles the postMessage round-trip. Used in Studio interact mode, Studio preview, V3 app preview, and any iframed embedding context where same-tab would hijack the parent.
- **Cordova native (iOS / Android)** — `cordova.InAppBrowser` opens the unified sign-in page. The widget intercepts IAB navigation against the sentinel callback URL `<api>/v1/auth/return-token?state=<nonce>` via both `loadstart` and `loadstop` (the latter is required on Android — `loadstart` doesn't always fire for same-origin `location.href` redirects).

## Why

- One sign-in surface across web, native, and SSO. The widget no longer carries its own password / 2FA / reset UI — the auth-loader handles those, including SAML2/OAuth round-trips.
- Removes ~460 lines of widget code that re-implemented flows the central auth-loader now owns.
- Closes the session-fixation surface the new same-tab flow opens up via a state-nonce round-trip + user-shape validation.

## Security model (same-tab and IAB return legs)

| Defence | Implementation |
|---|---|
| State nonce | `crypto.getRandomValues` → 128-bit hex, stored in `sessionStorage`, included in the callback URL. The auth-loader template preserves callback query params when it appends `&token=...&user=...`, so the state round-trips back to us. Single-shot — consumed on either success or rejection. |
| User-shape validation | Parsed `?user=` payload must be a plain object with numeric `id` and string `email`. Anything else → reject, scrub URL, surface error toast. |
| URL scrub | `cleanAuthReturnParamsFromUrl()` strips `token`/`user`/`state`/`error` between parsing and any storage write, so the token doesn't persist in address bar / history / bookmarks. |
| IAB callback match | Strict origin + pathname comparison via `new URL(event.url)` — not a prefix match. Closes look-alikes like `/v1/auth/return-token-anything?token=…`. |
| Console logging | IAB navigation logs mask `token`/`user`/`state` query params before output. |
| Fliplet.Auth missing | Popup path guards `Fliplet.Auth.signIn` availability; falls back to same-tab rather than throwing. |

## Removed UI surface (migrated to the central auth-loader)

- Email/password sign-in form
- Two-factor authentication input
- Forgot password / send reset code
- Password reset (new password + confirm)
- Forced password update (when org policy changes)
- All associated translation keys (`login.*`, `restore.*`, `verify.*`, `reset.*`, `twoFactor.*`, `update.*`, `infoToast.*`, `errorToast.*`, `successToast.passwordUpdated`)

If you were relying on any of these inline flows in a custom build, they now live on `/v1/auth/login` and the auth-loader bundle. The widget invokes that flow rather than rendering it.

## Companion / cross-team

- **API:** Fliplet/fliplet-api#8077 (DEV-1090) — the unified auth-loader, `/v1/auth/login` GET handler with `return=callback` / `return=postmessage` modes, the `fliplet_widget_login` HMAC-signed cookie for SAML2/OAuth round-trip context, and `/v1/auth/return-token` fallback page used by the IAB path.
- **SAML2/OAuth widgets:** Out of scope for this PR. `fliplet-widget-sso-saml2` and `fliplet-widget-sso-oauth2` still need a cross-team check that they handle the new SAML2 callback error shape (per Komsomol's note on #8077).

### Web — same-tab (deployed app on a real domain)

- [ ] Click **Sign in** on the widget. Browser navigates to `<api>/v1/auth/login?return=callback&callback=...&state=...`
- [ ] Complete sign-in. Browser returns to the original page with `?token=...&user=...&state=...`
- [ ] Widget signs the user in; URL is scrubbed; sign-in proceeds to the configured next page
- [ ] On bad return URL (no state / mismatched state / malformed user JSON / wrong user shape): widget shows error toast, URL is scrubbed, no session is written
- [ ] Manually constructing a URL with `?token=ATTACKER&user={"id":1,"email":"a@b"}` (no `state`) does NOT sign anyone in
- [ ] sessionStorage `fliplet_login_state` is cleared after the return is processed

### Web — popup (Studio interact / Studio preview / V3 app preview / iframed)

- [ ] In each context, clicking **Sign in** opens a popup (not a same-tab redirect)
- [ ] Successful sign-in via popup completes via postMessage and the widget transitions to the next page
- [ ] Cancelling the popup (closing it) does NOT surface an error toast
- [ ] Other popup errors DO surface the error message
- [ ] If `Fliplet.Auth.signIn` is unavailable (test by temporarily nulling `Fliplet.Auth` in DevTools before click): the widget falls back to same-tab without throwing

### Native — Cordova IAB

- [ ] **iOS:** Click **Sign in** → IAB opens the unified sign-in page → complete sign-in → IAB closes → widget signs the user in (via `loadstart`)
- [ ] **Android:** Same as above, exercised via the `loadstop` safety net (Android sometimes misses the final hop on `loadstart`)
- [ ] IAB DevTools console shows the navigation log with `token`/`user`/`state` masked as `<redacted>`
- [ ] Closing the IAB before completing returns the widget to the idle state (button re-enabled)
- [ ] On bad callback URL (no state / mismatched state / invalid user): IAB closes, widget shows error toast, no session written
- [ ] A look-alike callback URL like `/v1/auth/return-token-anything?token=...` is NOT accepted

### Regressions

- [ ] `Fliplet.Auth.signIn()` cancellation (user closes the popup) is still classified as a cancel — no toast
- [ ] The widget's "already signed in" detection (reads `fliplet_login_component` from App.Storage) still skips the button when a session is already restored
- [ ] Non-admin sign-in still surfaces `userNotAFlipletAdmin` (per `js/app.js`)

## Out of scope (deferred / documented)

- **Cross-origin iframe popup behaviour** (Arpan's W1): popups in cross-origin iframes are reliably blocked, but same-tab in an iframe is also broken UX. Needs a product decision on the embedded-app sign-in strategy.
- **Token in URL → analytics/Sentry/GTM URL snapshots** (Arpan's C2 secondary): only fully fixable by removing token-from-URL entirely (httpOnly cookie or postMessage even for same-tab). The state-nonce makes a replayed URL useless, which closes the realistic exploit path.
- **`Fliplet.Auth.signIn()` cancel detection via regex** (Arpan's W7): brittle but the SDK doesn't expose `err.code` today. Worth revisiting when the SDK formalises its error contract.
- **`var` → `let`/`const`** (Arpan's W6): file convention is `var`. No enforced lint rule.


Arpanexe:
## Summary
Refactors the Fliplet login widget from an inline form (email/password, 2FA, forgot-password, password reset, force-update-password) into a single "Sign in" button that delegates authentication to a unified auth-loader hosted on the Fliplet API. Three sign-in paths converge on a shared `handleAuthSuccess({ token, user })`: web same-tab (`window.location.assign` + return-leg query-string parsing), web popup (delegated to a new `Fliplet.Auth.signIn()` SDK call), and native Cordova (InAppBrowser with `loadstart`/`loadstop` URL interception against the API's `/v1/auth/return-token` callback). Net deletion of 466 lines (+531 / −997 across 7 files). Adds a configurable button label, a new `fliplet-auth` dependency, and removes all in-widget password / 2FA / reset UI.

## Critical Issues (Must Fix Before Merge)

1. **`build.js` `handleSameTabReturn`: same-tab return leg has no integrity check — session-fixation surface.** The function reads `token` and `user` directly from `Fliplet.Navigate.query` and pipes them into `Fliplet.User.setAuthToken(authResult.token)` + `Fliplet.Login.updateUserStorage(...)`. There is no `state` / `nonce` value sent on the outbound `buildSameTabCallbackUrl()` and validated on return, no origin check, no signature on the returning token. Anyone who can convince a victim to follow `https://victim-app.com/?token=ATTACKER&user={"id":1,"email":"a@b.com"}` will pin an attacker-controlled session onto the victim's local storage. Generate a one-shot `state` value before redirect (store in `sessionStorage`), pass it to the auth-loader, and reject the return-leg if `state` is missing/mismatched.

2. **`build.js` `cleanAuthReturnParamsFromUrl`: token leaks via URL, history, and referrer.** `cleanAuthReturnParamsFromUrl()` calls `history.replaceState` *after* the token has already been: parsed into `q.token`, written into local storage, set on `Fliplet.User`, and exposed in `window.location.href` for any inline-loaded analytics/error-tracking scripts (Sentry, GTM, Datadog RUM) that snapshot URLs at load time. Scrub the URL **before** any further work — call `cleanAuthReturnParamsFromUrl()` immediately after parsing the values into local vars, *then* call `handleAuthSuccess`. Better still: never use query-string for the token at all — POST the credentials from the auth-loader to a same-origin endpoint that sets an httpOnly cookie, or use the popup-postMessage path even for web.

3. **`build.js` same-tab and IAB paths: `JSON.parse(q.user || 'null')` with no schema validation.** `authResult.user.id`, `userRoleId`, `email`, `legacy` are written directly into `Fliplet.Login.updateUserStorage`. Combined with finding #1, an attacker can write arbitrary `{id, email, userRoleId}` into the victim's stored user record. Validate the parsed object shape (required fields, types) before passing it on, and treat anything else as a sign-in failure.

4. **PR has no description.** The body is empty. For a refactor that removes an entire password/2FA/reset flow, adds a new SDK dependency (`fliplet-auth`), introduces three new auth paths, and ships into `production`, the lack of a written intent makes the change very hard to audit and roll back. Add a Summary + Test plan covering all three flows (web same-tab, web popup, native IAB on iOS and Android) and the migration path for apps still relying on in-widget 2FA / forced-password-update.

5. **Codacy CI check is failing** (`ACTION_REQUIRED`). Pull and address the specific findings from https://app.codacy.com/gh/Fliplet/fliplet-widget-fliplet-login/pull-requests/98 before merging — even if it turns out to be a stylistic issue, a red status check on a release-train PR is a deploy-gate.

## Warnings (Should Address)

1. **`isStudioOrPreviewContext()` treats *every* cross-origin iframe as "use the popup path".** Many Fliplet apps are embedded on customer websites via iframe — those would previously have used the inline form and now will use `Fliplet.Auth.signIn()` which opens a popup; popups inside cross-origin iframes are reliably blocked by all modern browsers without explicit user gesture. Confirm what the desired UX is for customer-embedded apps; the current logic delivers a "popup blocked" error in exactly the case the PR was supposed to improve.

2. **`fliplet-auth` is a new build dependency** (`widget.json`). If the customer's app isn't built against a Fliplet runtime that ships `Fliplet.Auth`, the popup path will fail with `Cannot read properties of undefined (reading 'signIn')`. Confirm `Fliplet.Auth` is present in the runtime versions this widget targets, and add a defensive `if (!Fliplet.Auth || !Fliplet.Auth.signIn) { … }` guard around `openSignInPopup()`.

3. **`indexOf(callbackPrefix) === 0` is a fragile prefix match** (`build.js`, IAB handler). With `callbackPrefix = 'https://api.fliplet.com/v1/auth/return-token'`, a URL such as `https://api.fliplet.com/v1/auth/return-token-anything?token=x` matches. Use `event.url === callbackPrefix` or `event.url.startsWith(callbackPrefix + '?')`, or parse with `new URL(event.url)` and compare `origin + pathname` exactly.

4. **`console.log('[Fliplet.Login][IAB …]', event.url)` leaks the token to the console.** Tokens in screenshots/screen recordings shared in support tickets are a known leak vector; in some apps console output is captured by a remote log aggregator. Mask the `token`/`user` query params before logging, or log only the URL pathname.

5. **`parseQueryString()` re-implements `URLSearchParams`.** The same file already uses `new URL()`. The local implementation handles `+`/`%20` differently from the spec and won't tolerate hash fragments or repeated keys gracefully. Replace with `new URL(event.url).searchParams`.

6. **`var` usage throughout new code** — repo preference is `let`/`const` for new code even when surrounding file uses `var`. New declarations to convert in `build.js`: `var apiHost`, `var params`, `var appId`, `var url`, `var host`, `var match`, `var mode`, `var iabHandled`, `var browser`, `var qs`, `var user`, `var options`, `var loginUrl`, `var $btn`, `var original`, `var isWeb`, etc.

7. **`Fliplet.Auth.signIn()` cancel detection by regex on `err.message`** (`if (/cancelled|closed/i.test(message))`) is brittle. If the SDK changes wording (e.g., "user dismissed", "popup_closed_by_user") or localizes the message, every cancel becomes a visible error toast. Prefer `err.code === 'CANCELLED'` if the SDK surfaces one — or document the contract.

8. **Translation file (`translation.json`) keeps dead keys.** All the removed UI (`login.emailLabel`, `restore.*`, `verify.*`, `reset.*`, `twoFactor.*`, `update.*`, `infoToast.*`, `errorToast.*`, `successToast.passwordUpdated`) is still present. Strip them so future readers don't think the flow still exists.

9. **PR title (`release/DEV-1090`) and branch name (`projects/DEV-908-unified-login`) reference different tickets.** Aligns with the companion API PR #8077 (also DEV-1090), but the branch name says DEV-908. Worth picking one ticket and updating the other reference.

10. **Branch target is `production`, not `master`.** If this repo deliberately uses `production` as the integration branch, fine — but confirm against the team convention, and note that this means the PR will deploy directly without a staging hop.

## Suggestions (Nice to Have)

- Add Jest tests for the three sign-in paths (web same-tab, web popup, native IAB), plus negative tests for malformed `?user=` and missing `?token=` on the return leg.
- Add an integration test (or at minimum a manual checklist in the PR body) covering iOS IAB `loadstart` *and* Android IAB `loadstop` paths separately — the comment in `onLoadStop` notes Android needs the safety net, which means there's an existing platform-specific bug worth pinning down.
- Consider feature-flagging the new flow (`Fliplet.Env.get('unifiedLoginEnabled')`) so customers on legacy builds can fall back if `Fliplet.Auth` isn't present.
- `getApiHost()` returns `undefined` if neither `primaryApiUrl` nor `apiUrl` is set; `getApiOrigin()` then calls `new URL(undefined)`, catches, then `host.match(...)` throws on undefined. Guard with `var host = getApiHost() || ''` before the regex.
- The same-tab callback URL is built by hand in `buildCallbackLoginUrl` — prefer `new URL()` + `searchParams.set()` for consistency with the rest of the file.

## Verdict

**REQUEST_CHANGES.**

The same-tab return leg is the blocker: accepting an unsigned, untyped `?token=`/`?user=` from the URL and immediately stamping it into the session is a session-fixation primitive that anyone who can deliver a URL to a victim can exploit. That, plus the unvalidated `JSON.parse(q.user)`, plus the token leaking into the address bar / referrer / analytics URLs before scrubbing, are all in the same code path and all fixable in a single revision. Combined with the failing Codacy check, the empty PR description, and the open compat questions about `Fliplet.Auth` availability and iframe contexts, this isn't ready for production.

The refactor itself is a clear architectural win — removing the in-widget password UI in favor of the central auth-loader is the right direction, and the convergence on `handleAuthSuccess` is clean. Once the same-tab return is hardened with a `state`/`nonce` round-trip (or moved off query-string entirely), the user payload is schema-validated, the URL is scrubbed before any other code reads it, and the Codacy findings are addressed, this should be approvable.

Note: this widget is the client-side companion to the API PR #8077 — the Komsomol comment on that PR explicitly asked whether SSO consumer widgets handle the new SAML2 callback error shape. This PR doesn't address SAML2 at all (Fliplet email/password → unified auth-loader only), so that cross-team verification is still outstanding for `fliplet-widget-sso-saml2` / `fliplet-widget-sso-oauth2`.

---

*Posted by Claude Code on behalf of @ArpanJokhakar.*

hcarneiro:
Thanks for the review. Addressed in f3f0e82 — summary below.

## Applied

| # | Fix |
|---|---|
| C1 | State nonce round-trip via sessionStorage (single-shot, consumed on success or rejection) on both same-tab and IAB return legs. Echoed through the callback URL — the auth-loader template preserves callback query params when it appends `&token=...&user=...`. |
| C3 | `isValidUserShape(user)` guard requires `{id: Number, email: String}` before any storage write. |
| C4 | PR description filled in (Summary, security model, removed-UI list, test plan, deferred items). |
| W2 | `openSignInPopup` guards `Fliplet.Auth.signIn` availability; falls back to same-tab if missing. |
| W3 | IAB callback match is strict origin + pathname via `new URL()` — closes `/v1/auth/return-token-anything?token=…` look-alikes. |
| W4 | IAB `console.log` runs URLs through `maskUrlForLogging()` — `token`/`user`/`state` redacted. |
| W5 | Dropped bespoke `parseQueryString`; uses `new URL().searchParams`. |
| W8 | `translation.json` stripped to the 10 keys actually referenced by `build.html` + `js/{build,app}.js`. |

## Not applied (with reasons)

- **C2 secondary concern** — `cleanAuthReturnParamsFromUrl()` already runs before any storage write, so the "scrub after storage" claim doesn't apply. The remaining concern (third-party scripts snapshot `window.location.href` at load) is only fully fixable by removing token-from-URL entirely (httpOnly cookie or postMessage even for same-tab) — that's an API-side rewrite. The C1 state-nonce makes a replayed URL useless, which closes the realistic exploit path.
- **C5 (Codacy)** — Vendor-side infra issue, not a finding on the widget. Codacy's pipeline errors out on the Diff/Deltas step (zero-second elapsed, "An error occurred during this step") regardless of what's in the code. Not blocking.
- **W1** — Cross-origin iframe → popup → blocked needs a product call on the embedded-app sign-in strategy. Noted as deferred in the PR body.
- **W6** — `var` → `let`/`const`: file convention is `var`; no enforced lint rule. Mixing styles creates noise.
- **W7** — Cancel detection by regex is brittle, but `Fliplet.Auth.signIn` doesn't expose `err.code` today. Worth revisiting when the SDK formalises its error contract.
- **W9, W10, all suggestions** — Out of scope for this PR; some are worth follow-up tickets (Jest tests on the three paths, Android `loadstop` deep-dive).

Arpanexe (approved):
LGTM

PR #100, opened 2026-07-07
Revert "release/DEV-1090"

zeryabkhan91:
Reverts Fliplet/fliplet-widget-fliplet-login#98

### DIFF
--- a/build.html
+++ b/build.html
@@ -2,26 +2,342 @@
   <div class="login-loader-holder animated">
     <div class="loader" data-translate="widgets.login.fliplet.loading">Loading...</div>
   </div>
-  <div class="login-form-holder content-wrapper fliplet-login-button-wrapper">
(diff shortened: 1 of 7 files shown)


Komsomol (approved):
Approved. Clean GitHub-generated revert of #98 (DEV-1090 unified login): exact inverse line counts across the same 7 files, restores the pre-DEV-1090 in-widget email/password + 2FA + reset flows. No manual conflict edits to review.

Two things before merge, not blockers on the approval:
1. CI hasn't reported yet (mergeable_state unstable). Let it go green first.
2. Companion API side (fliplet-api #8077) still ships the `/v1/auth/login` unified auth-loader. Reverting the widget alone leaves the widget back on its own login while the API endpoint stays live. Confirm that's intended and no other app/widget depends on the widget invoking `/v1/auth/login`.

PR #101, opened 2026-07-08
DEV-1572: re-land unified login (#98) + passport-based session detection fixes

hcarneiro:
## Merge target

This PR merges into `projects/DEV-1572` (cut off `production`) so engineering can review here and QA can merge and test. **The final `projects/DEV-1572` → `production` release must NOT happen until Fliplet/fliplet-api#8331 is verified on staging and released to production** — the unified flow depends on the API creating the `flipletLogin` passport at login and returning the session token from the auto-success leg. Releasing the widget first reproduces the login loop that forced revert #100.

## Summary

Re-lands unified login and adds the session-detection fixes that were stranded on the orphaned `fix/PS-1956` branch when revert #100 shipped instead. Four commits:

1. **Reapply "release/DEV-1090"** — revert of revert #100; tree verified byte-identical to the original #98 merge.
2. **Passport used for session validation** (Zeryab, 11 Jun) — restores `getFlipletPassport()` passport-first detection in `initSession()`.
3. **Restore passport-based session detection** (Zeryab, 11 Jun) — validates stored tokens against `/v1/session`, rejects stale and app-token sessions instead of trusting App.Storage unconditionally.
4. **Reject app-token "logins" in `handleAuthSuccess`** (Zeryab, 11 Jun) — defense in depth for the native IAB shared-cookie-jar case; the primary guard is server-side (#8182, included in the API PR).

The diagnostic-logging commit from `fix/PS-1956` was deliberately dropped. Commits 2–4 preserve original authorship. Net vs production: 7 files, +838/−1056.

## Test plan (after the API fix reaches staging)

Per DEV-1572: password, 2FA, and SSO logins work; users from any region can log in to Fliplet Viewer; `flipletLogin` passport present after login and removed after logout. Run the Fliplet Viewer repro from PS-1956 on a **clean iOS build with the widget update confirmed applied** — the June iOS retests were contaminated by stale bundles. Also verify Studio preview signed-in detection and the Multi-Dev Environment dashboard.


zeryabkhan91:
## Review notes — paired with Fliplet/fliplet-api#8331

The passport-first detection here is the right structural fix: `getFlipletPassport()`, `getPassportForToken()` validating the stored token against `/v1/session`, and the app-token / empty-passport handling. `isValidUserShape()`, `isAppTokenUser()` and `maskUrlForLogging()` are welcome hardening. Two things I'd confirm before release:

### ⚠️ `js/app.js` is untouched — please confirm the logout-wipe cycle is fully closed
`verifyUserForDevEnvApp` and `getUserData()` still live in `js/app.js` unchanged (`js/app.js:111`, still called at `js/app.js:270`). `getUserData()` authenticates `GET /v1/user` with the App-Storage (`fliplet_login_component`) token and, **when that's empty, falls back to the app-token user (role ≠ admin) → `Fliplet.Session.logout()` + `App.Storage.remove(...)`** — a self-perpetuating logout/wipe.

Reproduced deterministically: empty `fliplet_login_component` → `GET /v1/user` resolves `{ id: 106, type: 'appToken', userRoleId: null }` → `verifyUserForDevEnvApp` rejects `USER_NOT_FLIPLET_ADMIN` → `logout()` + `remove('fliplet_login_component')` → stays empty → repeats (and rotates the app-cookie session, clearing the freshly-attached passport).

The new `build.js` detection stops `initSession` from reaching the gate with an app-token, but `validateAccount()` is also called directly (`js/app.js:211`, `:225`, `:307`, and `handleAuthSuccess`). Please confirm **every** `validateAccount` entry point now flows through the passport-first detection so the app-token fallback can't re-trigger the cycle — this is the exact failure that was reproduced against the dev-env dashboard.

### ⚠️ `isAppTokenUser()` email-pattern coverage
It matches `/^token-[\w-]+@fliplet\.com$/i`. App-token accounts also use `%web%` / `%app-list%`-style emails elsewhere on the platform, and the callback `user` payload carries no `type`, so please confirm the pattern covers all app-token email formats — otherwise a non-matching app token slips this client-side guard. (Lower severity: the server-side #8182 guard is the primary defence.)

hcarneiro:
@zeryabkhan91 both confirmed and fixed in 2551691 — the first one is the sharpest catch of this whole re-land; thank you.

**Logout-wipe cycle**: verified end to end, and it was live. The boot-time `validateAccount()` at `js/app.js:307` runs on every screen load via `Fliplet.Cache` and never passes through the new passport-first detection in `build.js`, so with empty storage `getUserData()` sent `GET /v1/user` with an undefined `Auth-token` header, fell back to the ambient app-token credentials, and the Dev-Env admin gate tore the session down exactly as you reproduced. Rather than patching the gate, the fix stops the class at the root: `getUserData()` now rejects with a coded `NO_STORED_SESSION` error **before** any request when storage has no token — no ambient fallback, no logout, no session rotation, on any entry point.

Entry-point audit you asked for:
- `build.js` `handleAuthSuccess` → `validateAccount({ data })` — data provided by the fresh login, never calls `getUserData()`.
- `build.js` `initSession` → `validateAccount({ updateUserStorage: true })` — only reachable after passport-first detection has populated storage with a verified token.
- `app.js:211/225` (account-setup IAB onclose) and `app.js:307` (boot) → `getUserData()` path, now guarded by `NO_STORED_SESSION`.
- Also swept `goToAccountSetup` (`app.js:204`), which sends `storage.auth_token` as a header too — only reachable after a successful validation, so storage is guaranteed populated.

One deliberate behavior change to be aware of: previously, a signed-out boot on a *non*-dev-env app would "validate" the app-token user and fire `flipletAccountValidated`. It now rejects silently instead — firing that hook for an app token was arguably a bug, but flagging it in case any consumer depended on it.

**`isAppTokenUser` pattern**: confirmed — token names are user-supplied (`apps-tokens.js` only replaces spaces), so punctuation like `token-123-prod-(eu)-7@fliplet.com` slipped `[\w-]`. Broadened to `/^token-[^@]*@fliplet\.com$/i`; false positives are implausible (`@fliplet.com` is staff-only and no human account starts with `token-`), and this stays tertiary behind the server-side guards (#8182 + the Studio-session/auto-success work in Fliplet/fliplet-api#8331).

zeryabkhan91:
### ⚠️ `debug is not defined` — ReferenceError on the app-token rejection path

`js/build.js:547` calls `debug('handleAuthSuccess: returned user is an APP TOKEN…', { … })`, but `debug` is defined nowhere in the widget — grepping the PR head, that's the only occurrence across `build.js`, `app.js`, and `interface.js`. The PR body notes *"the diagnostic-logging commit from `fix/PS-1956` was deliberately dropped"* — that commit is where `debug()` lived, and this one call site outlived it.

**Failure scenario:** it's on the app-token defence-in-depth branch of `handleAuthSuccess` — exactly the case this PR exists to harden (native IAB shares the app WebView cookie jar, so a stale/missing server guard can auto-complete the login as the app's bootstrap appToken). When `isAppTokenUser(authResult.user)` is true, execution hits `debug(...)` and throws **before** the `Fliplet.UI.Toast.error(...)` and `hideLoadingState()` below it ever run:

- popup flow (`Fliplet.Auth.signIn().then(handleAuthSuccess)`) → unhandled rejection, sign-in button stuck spinning
- same-tab / IAB flows → throw swallowed by the caller, same stuck-button result, no error toast

So the safety net silently breaks in the one scenario it was written for, instead of cleanly rejecting and keeping the user on the login form.

**Fix:** drop in the file's existing logging convention —

```js
console.warn('[Fliplet.Login] handleAuthSuccess: returned user is an app token, not a real login — rejecting', {
  userId: authResult.user.id,
  userEmail: authResult.user.email
});
```

— or restore a small `debug()` helper if you want to keep the call sites. Trivial change, but worth blocking on since it's on a security path.

hcarneiro:
@zeryabkhan91 confirmed and fixed in 2a90eaf. You traced the origin exactly right: `debug()` lived in the diagnostic-logging commit that was deliberately dropped from this re-land, and the app-token rejection branch (cherry-picked from a later commit that depended on it) kept the one dangling call site — so the guard threw a ReferenceError before the toast and button reset could run, in precisely the scenario it exists for.

Replaced with the file's `console.warn('[Fliplet.Login] …')` convention per your suggestion rather than restoring a `debug()` helper — it was a single call site, so the helper isn't warranted. Swept `build.js`, `app.js`, and `interface.js`: zero `debug(` references remain.

Mea culpa on process: when I dropped the diagnostics commit I checked that the remaining commits cherry-picked cleanly but didn't grep for references into the dropped code — that's exactly the kind of dangling dependency a clean cherry-pick can hide. Good block.

zeryabkhan91 (approved):
LGTM

PR #102, opened 2026-07-15
DEV-1572: attach passport to the app session + defer native success until IAB exit

hcarneiro:
### 1. Attach the flipletLogin passport to the app's CURRENT session
Root cause of "app list still shows Sign In" and protected screens bouncing after a successful unified login: the sign-in stores the passport on the **auth page's session** and hands the widget a token — but app list (`validateUserToken_`), protected-screen security, and every other consumer decide "signed in" by reading the **app's own session** via `getCachedSession()`, which additionally persists a snapshot with a 30-minute renewal. Nothing transferred the passport or refreshed that cache. The old inline flow got both for free because its login XHR ran in the app's context, so `passports.storeDetails()` wrote to the app's session directly.

Fix: after sign-in — and on the `initSession` restore path whenever the cached session lacks the passport — call `POST /v1/session/providers/copy` with the signed-in session's token (the endpoint FV/PS-1925 already uses), using the runtime's **default credentials** so the copy targets the session the app runs under right now (app sessions rotate when the served page is re-fetched — an earlier boot-captured-token design was disproven live for exactly that reason). Then `getCachedSession({ force: true })`. The restore-path variant makes this self-healing: any screen with the login widget re-attaches after a session rotation.

Verified end-to-end on a local app through the real publish pipeline (fresh browser, no interception): login → redirect → app-list renders the user's apps; session, cached session, and storage all confirmed carrying `flipletLogin`.

### 2. Defer the native success path until the IAB has fully closed
Root cause of "login completes on native but no redirect": `tryHandle` ran `handleAuthSuccess` immediately after `browser.close()`, so the `Navigate.to` at the end of the chain could fire while the iOS IAB dismissal transition was in flight — a WebView navigation issued during that transition is silently swallowed. Fix: stash the auth result and run it from the `exit` handler (fires after `close()` on both platforms), with a 3s fallback if a plugin quirk drops the event. **Needs verification on a physical iOS device.**

### Known limitations (documented, not blockers)
- Web popup path (`Fliplet.Auth.signIn`, Studio preview/interact contexts): the SDK swaps the token before `handleAuthSuccess`, so the copy may no-op there; the restore path heals it on the next login-screen load, and navigation is disabled in those contexts anyway.
- Cross-region users: the copy source session lives in the home-region DB and the copy 400s softly (PS-1925's known story) — the EU/US/CA staging matrix in DEV-1572 covers this.
- Logout: app-list's Sign out does NOT remove the flipletLogin passport from the app session (pre-existing; observed live). DEV-1572 requirement 3 needs a decision on where passport removal belongs.


hcarneiro:
Follow-up commit 02c1249: the restore-path restructure had left `authToken` referenced out of scope (runtime ReferenceError that killed the restore before the passport attach — caught by re-testing the rotation scenario, confirmed via eslint no-undef sweep). Also preserves the stored SESSION token on restore passes — the passport only carries the user token, and future re-attach passes need the session token as the copy source.

Full web matrix now verified with the final published build through the real publish pipeline (no interception):
- Fresh form login → passport on app session + cache, app list renders ✅
- Auto-success (ambient session re-login) → same ✅
- Restore after app-session rotation (cache-cleared page re-fetch) → initSession re-attaches, auto-navigates ✅

Remaining: physical iOS device run for the IAB exit-deferral fix.

PR #103, opened 2026-07-20
Release: release/DEV-1572 → production

hcarneiro:
## Release: `release/DEV-1572` → `production`

| Title | Ticket | PR |
|-------|--------|-----|
| Re-land unified login (#98) + passport-based session detection fixes | [DEV-1572](https://weboo.atlassian.net/browse/DEV-1572) | #101 |
| Attach passport to the app session + defer native success until IAB exit | [DEV-1572](https://weboo.atlassian.net/browse/DEV-1572) | #102 |

QA sign-off: DEV-1572 is **Ready for Release** (staging-verified on web + native, 17 Jul). `release/DEV-1572` is content-identical to the QA-tested `projects/DEV-1572`.

### Deployment instructions
1. **Merge only AFTER the API release ([fliplet-api#8385](https://github.com/Fliplet/fliplet-api/pull/8385)) is live in production and verified** — without the API fixes, this widget reproduces the login loop that forced revert #100.
2. Pre-deploy: confirm production has NO `com.fliplet.login` package version > 1.0.0 (higher versions silently shadow this release — the staging QA trap).
3. Deploy = publish the 1.0.0 widget package from this branch + app publish updates for consumers (Fliplet Viewer, Multi-Dev Environments dashboard).
4. Post-deploy smoke: FV repro on iOS + Android (fresh install → login → open My Apps → preview; logout → re-login → preview), Multi-Dev dashboard admin login, web app-list signed-in detection.
5. Rollback: republish the previous 1.0.0 package build + app updates (precedent: revert #100).
6. Known non-blocking item tracked on the ticket: app-list Sign out does not remove the flipletLogin passport from the app session (pre-existing) — requirement-3 decision pending.

[DEV-1572]: https://weboo.atlassian.net/browse/DEV-1572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DEV-1572]: https://weboo.atlassian.net/browse/DEV-1572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zeryabkhan91 (approved):
LGTM

js/build.js
```diff
@@ -1,48 +1,39 @@
...
-  var userPassword;
-
   document.addEventListener('offline', function() {
     _this.$container.addClass('login-offline');
     scheduleCheck();
   });
 
   if (Fliplet.Navigate.query.error) {
-    _this.$container.find('.login-error-holder').html(Fliplet.Navigate.query.error);
+    _this.$container.find('.login-error-holder').html(Fliplet.Navigate.query.error).addClass('show');
```
Komsomol: **Critical: DOM XSS sink on the login screen.** `Fliplet.Navigate.query.error` is a URL query parameter written into the DOM with `.html()`. A crafted link to any app's login page (`?error=`) executes attacker markup exactly where a phishing payload is most valuable. The sink predates this PR, but the new flow formalises `error` as an auth-return parameter and this line is modified in the diff. One-word fix:

```suggestion
      _this.$container.find('.login-error-holder').text(Fliplet.Navigate.query.error).addClass('show');
```

js/build.js
```diff
@@ -62,593 +53,944 @@ Fliplet.Widget.instance('login', function(data) {
...
+        action: 'login_fail'
+      });
 
-      Fliplet.User.getCachedSession()
-        .then(function(session) {
-          preserveSession = session;
+      var errorMessage = Fliplet.parseError(err, T('widgets.login.fliplet.errors.unableLogin'));
 
-          var passport = session && session.accounts && session.accounts.flipletLogin;
+      _this.$container.find('.login-error-holder').html('<p>' + errorMessage + '</p>').addClass('show');
```
Komsomol: **Warning: second HTML sink.** `errorMessage` comes from `Fliplet.parseError(err, ...)`, which can echo server-supplied error strings into `.html()`. Lower risk than the `query.error` sink above (not directly URL-controlled), same fix: build the `<p>` element and set its content with `.text()`.

js/build.js
```diff
@@ -62,593 +53,944 @@ Fliplet.Widget.instance('login', function(data) {
...
+
+    if (!isValidUserShape(user)) {
+      return reject('user payload failed shape validation');
+    }
+
+    cleanAuthReturnParamsFromUrl();
+
+    handleAuthSuccess({ token: q.token, user: user });
+
+    return true;
```
Komsomol, line 473: **Critical (decision needed): session token transits the URL on the default deployed-web sign-in path.** `handleSameTabReturn` receives the session token as a `?token=` query param on a top-level GET to the app host, so it lands in apps-host access logs, proxy logs, and a browser-history entry until `replaceState` cleanup runs. Mitigations are good (one-shot state nonce, cleanup, log masking, popup path uses postMessage) and this is the established auth-loader callback contract, but it is new exposure versus the pre-#98 production widget, whose login was an XHR POST. Under the DEV-1231 transport rules this needs an explicit accept-or-migrate decision recorded on DEV-1572; the follow-up path would be a one-shot state-token exchange (`POST /v1/session/authorize/state` pattern) for the same-tab leg.

Komsomol (changes requested):
Code review (via /fliplet-pr-review)

**Summary**: Widget half of the DEV-1572 unified-login re-land (release/DEV-1572 into production, re-shipping #101 + #102, content-identical to the QA-tested projects branch, verified by SHA). Replaces the inline login form with delegation to the unified auth-loader via web popup (postMessage), web same-tab redirect, and native Cordova IAB, plus passport-first signed-in detection, passport attach via POST /v1/session/providers/copy, app-token rejection guards, a state nonce, and the NO_STORED_SESSION guard that closes the dev-env logout-wipe cycle. The release engineering is genuinely well done: deploy-order contract with fliplet-api#8385 explicit in both directions, Zeryab's #101 findings all demonstrably fixed, and the staging shadow-package trap converted into a pre-deploy audit step.

**Regression risk**: High blast radius: this widget is the login gate for every app using Fliplet login, and it is a re-land of code already reverted once (#98 reverted by #100) for a production login loop.

**Blockers** (inline comments on the code ones):
1. **CI is not green**: Codacy Static Code Analysis is action_required on head 101d55a, the only check on the PR. A production release PR should not merge over a non-green check; resolve or explicitly disposition the findings.
2. **XSS sink at js/build.js:31** (inline, one-word fix).
3. **Token-in-URL on the same-tab flow** (inline): needs an explicit accept-or-migrate decision under DEV-1231 recorded on the ticket.
4. **Ordering**: fliplet-api#8385 must be merged, deployed, and verified in production before this merges, per both PR bodies, including the prod widgets-table audit for a shadowing com.fliplet.login version > 1.0.0.

**Also address**:
- AC 3 on DEV-1572 (logout removes the passport) is a documented miss; needs Farhad's recorded acceptance on the ticket.
- #102's passport-copy and IAB-deferral logic reached this release without a prior engineering approval (merged by QA); the review on this PR should cover those js/build.js sections specifically, not just release mechanics.
- No automated tests across ~950 lines of new auth logic; at minimum the pure helpers (isAppTokenUser, isValidUserShape, getFlipletPassport, buildCallbackLoginUrl) are trivially unit-testable.

**Suggestions**: clear the 3s IAB fallback timer in onExit to avoid a rare race with the deferral fix; gate the per-navigation IAB console.log behind a debug flag after the rollout watch closes; fail closed instead of falling back to Math.random in generateAuthState.

**CI**: Red (Codacy: action_required).

hcarneiro:
@Komsomol code-level findings addressed in 098bcf1:

- **XSS sink at js/build.js:31** — fixed with your one-word `.text()` suggestion. Swept the branch for the class: exactly two `.html()` sinks existed (this one and :660), zero remain.
- **Second sink (:660)** — fixed by building the `<p>` element and setting content via `.text()`.
- **3s IAB fallback timer** — now cleared in `onExit` (double-run was already impossible via the `pendingAuthResult` null-check, but a stray callback shouldn't outlive the normal path).

Codacy re-run/disposition is being handled by Hugo. The two decision-records (token-in-URL accept-or-migrate under DEV-1231, AC3 logout acceptance) and the follow-up tickets (same-tab state-token exchange — the widget already uses the `POST /v1/session/authorize/state` pattern in `goToAccountSetup`, so the migration path is concrete; unit-test harness for the pure helpers) are being recorded on DEV-1572 separately.

On `crypto.getRandomValues` fail-closed: declining for now — support is universal in target browsers, and bricking login on an exotic webview to avoid a weaker-nonce fallback seems the worse trade given the nonce is defense-in-depth. Open to revisiting.

Note for QA: this commit touches only the two error-render statements and the timer — no flow changes — so a login-error smoke (bad password shows the message as plain text) should cover it without a full re-run.

hcarneiro:
New commit `dc8701e` — the token-in-URL finding is now **addressed**, not accepted: the widget sends `authExchange=1` and both return legs (same-tab + native IAB) receive a single-use `authState` instead of `token+user`, swapping it for the session via `GET /v1/session?state=`. The IAB pending result became a promise so the exchange composes with the exit-deferral (consumed exactly once by exit or the fallback timer). Legacy contract kept for rollout skew. Pairs with fliplet-api#8385 commits `1576637eb`+`57855b548` — deploy order unchanged (API first). QA re-run on staging should confirm the return URL shows `authState=` and never `token=`.

zeryabkhan91 (approved):

Arpanexe:
Review of current head `a63ddb0` — two issues remain:

1. **[P2] Same-tab authentication failures leave the login form hidden indefinitely** (`js/build.js:1113`).

   `handleSameTabReturn()` makes `init()` return before `initSession()` can reveal the UI. Its rejection paths only show a toast or reset the still-hidden button, so an expired/failed `authState` exchange, app-token rejection, or downstream login failure leaves the user staring at the loader until they manually reload. Ensure every same-tab failure calls `showStart()` (or otherwise reveals the form and enables retry).

2. **[P2] The configurable sign-in button label is overwritten by translation** (`build.html:17`).

   The unconditional `data-translate="widgets.login.fliplet.actions.signIn"` replaces the rendered `{{buttonLabel}}` when `translate()` runs, so Studio’s new “Sign-in button label” setting has no runtime effect. Apply the translation key only when `buttonLabel` is absent, or omit the translation attribute for customized labels.

Validation:
- JavaScript syntax checks passed and `translation.json` parses.
- Codacy remains `action_required`, leaving the PR merge state unstable.
- The required API dependency, Fliplet/fliplet-api#8385, is still open, so the documented API-first release condition is not yet satisfied.

hcarneiro:
@arpan both P2s fixed in `fa0a07b`:

1. **Same-tab failures leaving the form hidden** — `showStart()` now runs on every same-tab failure sink: the `reject()` choke point in `handleSameTabReturn` (covers nonce mismatch, failed/expired authState exchange, and user-shape rejection), the app-token guard, and the `handleAuthSuccess` catch (downstream login failure). It's a no-op in popup/IAB where the form is already visible.

2. **Custom button label overwritten by translation** — `data-translate` on the sign-in button is now emitted only when no `buttonLabel` is set (`{{#unless buttonLabel}}`), so Studio's Sign-in button label setting survives `translate()`.

On the remaining validation notes: Codacy `action_required` and the API-first dependency (#8385) are both being tracked by Hugo — #8385 is CI-green and pending its second approval; this widget PR merges only after it's live, per the documented order.

Holding scope here: any further adjacent findings we'll take as follow-up tickets (like DEV-1625 for the logout/passport removal) rather than growing this PR — the login-loop fix and the token-in-URL hardening are both now addressed.

js/build.js
```diff
@@ -62,593 +55,1079 @@ Fliplet.Widget.instance('login', function(data) {
...
+    if (!authState) {
+      return Promise.resolve(null);
+    }
 
-        Fliplet.UI.Toast(T('widgets.login.fliplet.successToast.passwordUpdated'));
+    // Absolute URL when authHost is known (Fliplet.API.request only
+    // prepends the app host to relative paths); relative otherwise.
+    var path = 'v1/session?state=' + encodeURIComponent(authState);
+    var url = authHost
+      ? String(authHost).replace(/\/+$/, '') + '/' + path
```
zeryabkhan91, line 372: **⚠️ Blocking — the cross-region fix doesn't work: `Fliplet.API.request` prepends the app's API host to *every* URL, absolute ones included.**

The comment above states the premise this whole approach rests on:

> `Fliplet.API.request` only prepends the app host to relative paths

That isn't the case. Checked against **live production** core (`cdn.api.fliplet.com/assets/fliplet-core/1.0/core.js:1671-1688`) — there's no absolute-URL branch:

```js
let apiUrl = options.apiUrl || Fliplet.Env.get('apiUrl');

if (apiUrl && apiUrl[apiUrl.length - 1] !== '/') apiUrl += '/';

options.url = apiUrl + options.url;   // unconditional
```

And the API sends an absolute origin — fliplet-api#8385 uses `exchangeBase = location.origin` on the auto-success leg and `result.user.host` on the 2FA leg. So the request that actually goes out is:

```
https://api.fliplet.com/https://us-api.fliplet.com/v1/session?state=…
```

which 404s, gets swallowed by the `.catch()` below, returns `null`, and surfaces as `'authState exchange failed or returned an invalid user'` → the "unable to login" toast. That's both legs (same-tab return and the native IAB path), and it fails for exactly the cross-region users this commit is meant to unblock.

The supported way to retarget the host is the `apiUrl` option (`core.js:1681`), which also normalises the trailing slash, so the `replace(/\/+$/, '')` handling can go too:

```js
var opts = { url: 'v1/session?state=' + encodeURIComponent(authState) };

if (authHost) {
  opts.apiUrl = authHost;
}

return Fliplet.API.request(opts).then(function(response) { … });
```

Worth confirming the core bundled into the shipped native apps supports `options.apiUrl` (production core does). If an older bundle ignores it, behaviour degrades to same-region — no worse than today, unlike the current malformed URL.

One follow-on to verify once the host is genuinely honoured: the exchange will carry `Authorization`, `X-Device-Platform`, `X-Device-Tracking` and `X-App-Locales`, so it triggers a CORS preflight against the *other region's* API host. Please confirm allow-origin/allow-headers are right there before calling this fixed — that config has drifted before.

js/build.js
```diff
@@ -62,593 +55,1079 @@ Fliplet.Widget.instance('login', function(data) {
...
+      return reject('state nonce missing or mismatch');
+    }
 
-      $('.help-two-factor').addClass('hidden');
-      _that.addClass('disabled');
-      _that.html(LABELS.sendProcessing);
+    // Preferred contract: one-shot state exchange — no credentials in the URL.
+    if (q.authState) {
+      var authState = q.authState;
+      var authHost = q.authHost;
```
zeryabkhan91, line 438: **⚠️ Blocking — `authHost` is an unvalidated, URL-supplied request target.**

On the same-tab leg `authHost` comes straight off the page's own query string and is used as the host the exchange is sent to, with no allowlist. Once the host is actually honoured (see the `exchangeAuthState` comment), that gives an attacker-chosen value two ways to bite:

1. **Credential exfiltration** — `Fliplet.API.request` attaches `Authorization: Bearer <base64 auth token>` to whatever host it's pointed at (`core.js:1694-1707`).
2. **Session fixation** — the response is trusted wholesale: `session.auth_token` and `session.user` flow into `handleAuthSuccess`, and `isValidUserShape` validates *shape*, not authenticity. The victim ends up signed in under a token the attacker controls.

The `state` nonce check above does gate this and the nonce is unguessable, so it isn't trivially exploitable. But that nonce transits URLs to the auth page — referrer headers, apps-host access logs — which is the exact leak class DEV-1231 exists to close, so it shouldn't be the only thing standing between a query param and an arbitrary request target.

The IAB leg is much lower risk, since its `authHost` is read from a URL already pinned to `expectedOrigin`.

Cheap defence in depth — validate before use, and fall back to the app host otherwise:

```js
function safeAuthHost(value) {
  if (!value) return null;

  try {
    var u = new URL(value);

    if (u.protocol !== 'https:') return null;

    return /(^|\.)fliplet\.com$/i.test(u.hostname) ? u.origin : null;
  } catch (err) {
    return null;
  }
}
```

(plus whatever dev-env hosts you need on the allowlist).

Minor, while you're in here: the commit message for `a63ddb0d` says `authHost` is "masked in IAB logs alongside the other credentials", but `maskUrlForLogging` only masks `['token', 'user', 'state', 'authState']`. Harmless — a hostname isn't a credential — but the claim will get trusted later. Stripping it from the URL/history is correctly done at both sites.

zeryabkhan91:
**⚠️ Blocking — the "content-identical to the QA-tested branch" claim in the PR body no longer holds.**

```
gh api repos/Fliplet/fliplet-widget-fliplet-login/compare/projects/DEV-1572...release/DEV-1572
→ ahead_by: 3
```

Three commits exist only on `release/DEV-1572` — they map to this PR and to no feature PR, so they were neither reviewed in #101/#102 nor part of the branch QA signed off:

| Commit | Date | What it does |
|---|---|---|
| `098bcf16` | 20 Jul | XSS sinks (`.html()` → `.text()`) + IAB fallback timer |
| `dc8701e3` | 20 Jul | **Replaces the entire auth return contract** with the one-shot `authState` exchange |
| `a63ddb0d` | 22 Jul | Changes *where* that state token is redeemed (`authHost`) |

(and `fa0a07b5`, 22 Jul, adds the same-tab failure form-reveal.)

The body still reads *"QA sign-off: DEV-1572 is Ready for Release (staging-verified on web + native, 17 Jul). `release/DEV-1572` is content-identical to the QA-tested `projects/DEV-1572`."* The 17 Jul sign-off predates all of the above, and `dc8701e3` changed the sign-in return leg wholesale — the thing QA actually exercised is not what would ship.

Could you update the body to list these commits explicitly, and re-run the return-leg QA across web + native × same-region + cross-region? Given the finding on `exchangeAuthState`, the cross-region case in particular has never passed.

Arpanexe:
## Re-review recap (final head `fa0a07b`)

The two issues from my previous review were fixed:

- Same-tab authentication failures now reveal the login form for retry.
- Custom sign-in button labels are no longer overwritten by translation.

Two blocking code issues remained on the final head:

1. **[P1] Cross-region exchange requests are malformed.** `exchangeAuthState()` passes an absolute URL to `Fliplet.API.request`, but the production Fliplet runtime unconditionally prepends its configured API host. This produces a URL such as `https://api.fliplet.com/https://us-api.fliplet.com/v1/session?state=…` and breaks both web and IAB cross-region login. The supported pattern is a relative `url` plus `apiUrl: authHost`. See the [existing inline finding](https://github.com/Fliplet/fliplet-widget-fliplet-login/pull/103#discussion_r3629696097).

2. **[P1] `authHost` is accepted directly from the return URL.** Once the request is correctly retargeted through `apiUrl`, `Fliplet.API.request` attaches authentication headers to that host. The value must be normalized and restricted to HTTPS allowlisted Fliplet/environment API origins before use. See the [existing inline finding](https://github.com/Fliplet/fliplet-widget-fliplet-login/pull/103#discussion_r3629698098).

Release validation also remained incomplete: the release branch was four commits ahead of the branch covered by the stated QA sign-off, including substantial changes to the authentication return flow; the return-leg matrix needed rerunning across web/native and same-region/cross-region. Codacy was `action_required` at review time, and the required API PR Fliplet/fliplet-api#8385 was not merged.

**Verdict at final head:** REQUEST_CHANGES.

_Note: PR #103 and its API dependency are now closed without merge; this comment records the final re-review state._

PR #104, opened 2026-07-22
DEV-1572: unified login re-land (clean, no token-in-URL exchange)

hcarneiro:
## Summary

Clean reconstruction of the DEV-1572 widget work off `production`, **with the token-in-URL `authState` exchange removed** (deferred to its own hardening ticket after review found it broken cross-region and an unvalidated request target). This is the QA-verified core plus the two safe review fixes — nothing that changes the login flow QA already signed off.

Replaces the previous release/DEV-1572 line (#103), which carried the exchange.

## Commits (cherry-picked, exchange skipped)

1. Re-land unified login (#98) + passport-based session detection (#101).
2. Attach passport to the app session + defer native success until IAB exit (#102). **(These two = the QA-verified core.)**
3. XSS sink fixes — both `.html()` error renders → `.text()` (reflected-XSS + parseError injection).
4. Reveal the login form on same-tab failures + honour a Studio-configured custom button label.

## What's NOT in it (deferred)

The `authState` exchange (`dc8701e`, `a63ddb0`). The auth return leg keeps the `token+user` contract, guarded by the state nonce + URL scrub + log masking — the state QA verified on web + native.

## Verification

- No exchange residue (`authState`/`authExchange`/`authHost` = 0); both error sinks are `.text()`; button-label honoured via `{{#unless buttonLabel}}`; form-reveal on all three same-tab failure paths.
- JS syntax + eslint clean.
- API contract confirmed against `fix/DEV-1572-2`: callback returns `token`+`user`, session carries the `flipletLogin` passport.

## Deploy order

Merge/release **only after** the API (Fliplet/fliplet-api#8406) is live and verified. Before publishing, confirm prod has no `com.fliplet.login` version > 1.0.0.


zeryabkhan91:
Reviewed alongside Fliplet/fliplet-api#8406. Dropping the `authState` exchange removed both blocking findings from #103, and what's left is the QA-verified shape plus the two safe fixes — nothing blocking here.

### Verified
- **No exchange residue.** The 9 remaining `authState` matches are all the state-*nonce* helpers (`generateAuthState` / `storeAuthState` / `consumeAuthState`); the `authState`/`authHost` query params and `exchangeAuthState()` are gone from both return legs.
- **Both XSS sinks are `.text()`** — the URL query param at `js/build.js:33`, and `parseError` at `:678-681` via `$('<p>').text(...)` rather than an HTML string.
- **IAB fallback timer is cleared on exit** (`:546-549`), and `pendingAuthResult` is consumed exactly once by whichever of `onExit`/the timer gets there first.
- **`showStart()` on all three same-tab failure sinks** — the `reject()` choke point (`:366`), the app-token guard (`:605`), and the `handleAuthSuccess` catch (`:685`). That's the right set: the same-tab path returns from `init()` before `initSession()` would reveal the form.
- Strict origin + pathname match on the IAB sentinel (`:441`) and the `primaryApiUrl` preference in `getApiHost()` (`:62-66`) are both retained.

### Suggestions (non-blocking)

**`rejectIab` doesn't call `showStart()`** (`:481-486`) while all three same-tab sinks do. That's correct today — the IAB path never hides the form, so `hideLoadingState()` is enough — but the asymmetry looks like an oversight next to the three sites that do call it. A one-line comment would stop someone "fixing" it later.

**`handleSameTabReturn` returns early on `!q.token`** (`:351`), so a return leg that arrives carrying only `?error=` never consumes the stored nonce — it lingers in sessionStorage until the next sign-in attempt overwrites it. Harmless in practice (it's still single-use against the next success), but it slightly undercuts the "burn the nonce on first arrival" property the comment just below describes.

**Still no automated tests.** The security-critical pieces here are pure functions — the nonce comparison, `isValidUserShape`, `isAppTokenUser` — so they're cheap to cover, and they're exactly the parts where a regression is silent. Given this widget has now been reverted once and re-landed twice, a small unit suite would pay for itself.

### Deploy
API-before-widget is right, and the "no `com.fliplet.login` > 1.0.0 in prod" pre-check is worth keeping prominent — that shadowing trap is what made staging QA misleading last time. Since both PRs target `projects/DEV-1572-2`, there's still a release PR to come; worth diffing its commit list against this one before sign-off rather than relying on a "content-identical" claim.

Arpanexe:
## Review findings

I found four issues that should be addressed before merging:

1. **[P1] Native session token is overwritten during restore** — [`js/build.js:944`](https://github.com/Fliplet/fliplet-widget-fliplet-login/blob/5f27f03f321937e4fe50ebcc05e74cd6e6490a31/js/build.js#L944)

   `initSession()` deliberately preserves the session token, then `validateAccount({ updateUserStorage: true })` replaces it with `/v1/user`'s bare user token. That token has no session/passport, so a subsequent native restore treats the user as signed out. Validate without rewriting storage, or explicitly preserve the existing session token.

2. **[P1] Account-setup requirements are silently bypassed after login** — [`js/build.js:644`](https://github.com/Fliplet/fliplet-widget-fliplet-login/blob/5f27f03f321937e4fe50ebcc05e74cd6e6490a31/js/build.js#L644)

   `authResult` is `{ token, user }`, while password-policy, agreement, profile, and 2FA setup flags are inside `authResult.user`. Passing the wrapper to `validateAccount()` makes `userMustSetupAccount()` see none of those flags, yet it still emits `flipletAccountValidated` and caches success. Fetching complete user data with `validateAccount()` would avoid both this and finding 1.

3. **[P1] Same-tab login fails on URLs with fragments/hash routes** — [`js/build.js:91`](https://github.com/Fliplet/fliplet-widget-fliplet-login/blob/5f27f03f321937e4fe50ebcc05e74cd6e6490a31/js/build.js#L91)

   The state is appended to the end of the callback string. For `https://app/#/route`, the result is `https://app/#/route?state=...`; the API then appends `token` and `user` inside the fragment too. `Fliplet.Navigate.query` never receives them, breaking published hash-routed apps. Insert callback parameters before the fragment or store and restore the fragment separately.

4. **[P2] Successful popup login corrupts Fliplet.Auth's stored-user record** — [`js/build.js:630`](https://github.com/Fliplet/fliplet-widget-fliplet-login/blob/5f27f03f321937e4fe50ebcc05e74cd6e6490a31/js/build.js#L630)

   `Fliplet.Auth.signIn()` stores `id`, `region`, and `legacy`, but the following `updateUserStorage()` replaces that object with only `userRoleId`, `auth_token`, and `email`. Consequently, `Fliplet.Auth.currentUser()` returns missing identity fields. Extend `updateUserStorage()` to retain the SDK's complete schema.

Syntax parsing and translation-key checks passed. GitHub currently reports no automated checks for this PR.

hcarneiro:
@arpan review findings addressed in `78742ad`:

- **F1 (native session token overwritten)** — `initSession` no longer passes `updateUserStorage: true` to `validateAccount`; it already wrote storage with the session token above, and the re-write was clobbering it with `/v1/user`'s bare user token. Fixed.
- **F2 (account-setup bypassed)** — `handleAuthSuccess` now passes `authResult.user` (which carries the 2FA/agreements/profile/password flags) to `validateAccount`, not the `{ token, user }` wrapper that hid them.
- **F3 (hash-routed same-tab break)** — the callback now inserts `state` **before** any `#fragment`; paired with the API fix (#8406) which appends `token`/`user` before the fragment too. Hash-routed (V3) apps now receive them in the query string.
- **Zeryab's `rejectIab` asymmetry** — added a one-line comment explaining why it intentionally omits `showStart()` (the IAB path never hides the form).

**F4 (`Fliplet.Auth.currentUser()` missing fields, P2)** — this one I want to get right rather than guess: `updateUserStorage` writes `fliplet_login_component` + `Fliplet.Profile`, and which fields `currentUser()` needs depends on the fliplet-core Auth/Profile schema that isn't in this repo. Rather than risk corrupting Profile storage with a blind fix, I'd like to confirm the exact schema first — proposing this as a small follow-up (it's P2, no login-flow impact). Happy to pair on it or take a pointer to the SDK's expected shape.

Deploy order unchanged (API #8406 first). Re-QA will exercise the hash-routed and setup-required paths these fixes touch.

js/build.js
```diff
@@ -62,593 +55,984 @@ Fliplet.Widget.instance('login', function(data) {
...
+      return Fliplet.Hooks.run('login', {
+        passport: 'fliplet',
+        userProfile: authResult.user
+      });
+    }).then(function() {
+      // Pass authResult.user, not the { token, user } wrapper: the account-setup
+      // flags (2FA link, agreements, profile, password-must-change) live on the
+      // user object, and userMustSetupAccount reads them off the top level — the
+      // wrapper hides them, silently skipping required setup.
+      return Fliplet.Login.validateAccount({ data: authResult.user });
```
zeryabkhan91: **⚠️ Blocking — this disables the Dev Environment admin gate, and doesn't deliver the setup flags it's meant to.**

`validateAccount` treats `options.data` as the **`getUserData()` envelope**, not as a user object (`js/app.js:283-291`):

```js
return getData.then(function(response) {
  return verifyUserForDevEnvApp(_.get(response, 'user')).then(…)   // reads .user off the envelope
}).then(function(response) {
  return userMustSetupAccount(response)                            // reads the flags off the top level
```

And `verifyUserForDevEnvApp` early-returns when handed nothing (`js/app.js:130`):

```js
if (!isDevEnvApp || !user || user.userRoleId === FLIPLET_ADMIN_ROLE_ID) {
  return Promise.resolve();
}
```

Passing `authResult.user` (a bare user) makes `_.get(response, 'user')` **`undefined`**, so the gate resolves immediately and does nothing. With the previous `{ token, user }` wrapper it at least had a `.user` to check, so the gate ran. Net effect: a non-Fliplet-admin signing in to a Dev Environment app is no longer logged out and rejected with `USER_NOT_FLIPLET_ADMIN` — and "Multi-Dev dashboard admin login" is on this release's own post-deploy smoke list.

The stated goal isn't reached either way. The callback user payload is the minimal `{ id, email, userRoleId, firstName, lastName, legacy }` shape (that's what `isValidUserShape` enforces), so it carries none of `mustLinkTwoFactor` / `mustReviewAgreements` / `mustUpdateProfile` / `policy` / `credentialTypes` that `userMustSetupAccount` reads. Both the old and the new shape produce "no setup required".

Suggest dropping `data` entirely:

```js
return Fliplet.Login.validateAccount();
```

`updateUserStorage` has already run at `:640` with the session token, so `getUserData()` will fetch `/v1/user` and return the real envelope — the setup flags actually populated, *and* `verifyUserForDevEnvApp` receiving a real user.

That's exactly what the sibling change in this same commit does for the restore path (`:962`), and for the same underlying reason. Right now the two halves of the commit disagree: one stops passing pre-fetched data so the real envelope is used, the other passes an even narrower object.

(The restore-path change itself is correct — `updateUserStorage` inside `validateAccount` writes `response.user.auth_token`, the bare user token, which would clobber the session token `initSession` just stored.)

zeryabkhan91:
Re-reviewed after `78742ad0`, alongside Fliplet/fliplet-api#8406.

**One blocking item**, inline on `js/build.js:658`: switching `validateAccount({ data: authResult })` to `{ data: authResult.user }` makes `_.get(response, 'user')` undefined, so `verifyUserForDevEnvApp` early-returns and the **Dev Environment admin gate no longer runs** on the fresh-login path — while still not surfacing the account-setup flags it was aiming at, because the callback user payload doesn't carry them. `validateAccount()` with no arguments does both correctly. Full reasoning in the inline comment.

Everything else in this commit looks right:

- **The restore-path change** (`validateAccount({ updateUserStorage: true })` → `validateAccount()`, `:962`) is correct and well explained — `updateUserStorage` inside `validateAccount` writes `response.user.auth_token`, the bare user token, which would clobber the session token `initSession` just stored and leave the next native restore signed out.
- **The hash-fragment fix** in `buildCallbackLoginUrl` is correct: `state` goes before any `#`, so on a hash-routed app it lands in the query string where `Fliplet.Navigate.query` can actually read it, and the API's matching `login.pug` change keeps `token`/`user` there too. Callbacks without a fragment are byte-identical to before.
- **The `rejectIab` comment** resolves the asymmetry I raised — thanks.

Still no automated tests in this widget. That's what let the `validateAccount` shape change through: the nonce comparison, `isValidUserShape`, `isAppTokenUser` and the `validateAccount` call shape are all cheap to cover, and they're precisely the places where a regression is silent rather than loud. Given this widget has now been reverted once and re-landed twice, a small unit suite would pay for itself quickly.

hcarneiro:
@zeryabkhan91 blocking item fixed in `6268547` — good catch, that was a real regression I introduced.

Switched `handleAuthSuccess` to `Fliplet.Login.validateAccount()` with no args (matching the restore-path change in the same commit). My earlier `{ data: authResult.user }` made `_.get(response, 'user')` undefined, so `verifyUserForDevEnvApp` early-returned and the Dev-Environment admin gate stopped rejecting non-admins — and it still didn't surface the setup flags (the callback payload doesn't carry them). With no args, `getUserData()` fetches `/v1/user` using the session token `updateUserStorage` already stored: the gate runs and the flags are populated.

On the no-unit-tests point: fair, and it's what let the shape change through. The widget has no test harness today; standing one up for the pure helpers (nonce compare, `isValidUserShape`, `isAppTokenUser`, the `validateAccount` call shape) is worth doing but is its own bit of infra — proposing it as a follow-up rather than growing this PR.

zeryabkhan91 (approved):
LGTM

PR #106, opened 2026-08-04
DEV-1633: redeem authState on its issuing host + validate authHost

zeryabkhan91:
### Product areas affected
Fliplet Login widget (`js/build.js`) — the sign-in return leg on both web same-tab and native in-app-browser. No interface/build config changes.

### What does this PR do?
Reinstates the one-shot `authState` exchange stripped from the DEV-1572 release (PR #103, commits `dc8701e` / `a63ddb0`) and fixes the two blocking review findings that caused the revert. The widget sends `authExchange=1` on the login URL, then swaps the returned `authState` for the session via `GET /v1/session?state=` — so neither the session token nor the user payload ever transits a URL.

Paired with Fliplet/fliplet-api#8465 — **neither PR works alone.** Deploy the API first.

### JIRA ticket
[DEV-1633](https://weboo.atlassian.net/browse/DEV-1633)

### Result
Verified end-to-end in a real browser on a published local web app: sign-in completes, the callback carries `authState`/`authHost` and never `token=`/`user=`, the URL is scrubbed, and the `flipletLogin` passport is attached. A genuine cross-region login (US-only user against an EU app) returned `200` with the CORS preflight passing `204`.

**Review findings fixed:**

1. **Cross-region redemption (was BLOCKING, functional).** The reverted code built an absolute URL, but `Fliplet.API.request` prepends the API base to *every* url — absolute ones included (`core.js`: `options.url = apiUrl + options.url`, no scheme guard) — producing `https://api.fliplet.com/https://us-api.fliplet.com/…` → 404. So the cross-region case never worked. Now uses the supported `options.apiUrl` retarget (DEV-667). The ticket cited `core.js:1671-1688`; the actual code is at **1767-1784** (that range is `hashString`/`hashCode`) — the older line numbers appear to be from a pre-DEV-667 revision.

2. **`authHost` was an unvalidated request target (was BLOCKING, security).** On the same-tab leg it comes straight off the page query string and would become the host receiving `Authorization: Bearer`, with the response fed into `handleAuthSuccess`, which shape-validates the user but not its authenticity — credential exfiltration plus session fixation from a crafted link. `safeAuthHost()` now gates it.

3. **CORS (verify).** No change needed. `config.allowedOrigins` already includes every region's `host`/`apps_host`/`cdn_host`, and `libs/cors.js` sets no `allowedHeaders`, so the package reflects `Access-Control-Request-Headers`. Confirmed live: cross-region preflight returns `204` with all four custom headers reflected; a hostile origin gets no CORS headers.

Also adds `authHost` to `maskUrlForLogging` (not a credential, but masking it stops a log line revealing which region a user's session lives in).

### Checklist
 - [x] Added automated test coverage as appropriate for this change.

`safeAuthHost` is now in its own file (`js/safe-auth-host.js`) taking the app origin as an argument, which makes it a pure function testable outside a browser — `js/build.js` calls `Fliplet.Widget.instance` at module scope and cannot be loaded in Node.

**18 tests, `npm test` → `node --test`, no new dependency.** They cover the hostile inputs that previously only lived in a throwaway harness (suffix spoofs, query-string spoofs, userinfo `@` tricks, `javascript:`, protocol-relative, missing label boundaries) plus everything the two review rounds surfaced: the production CDN origins, sibling-rule depth, ports, and trailing-dot FQDNs.

Note the widget has **no CI running these** — Codacy is the only check and it reports `action_required` on every PR in this repo (#99 through #106 identical), which is an integration-authorisation state, not a finding against this branch. Wiring up a workflow to run `npm test` is worth doing but is arguably separate from this PR.

### Deployment instructions
Deploy Fliplet/fliplet-api#8465 **first**, then publish this widget. An old API ignores `authExchange=1` and keeps the legacy `token`+`user` contract, so the ordering is safe in both directions.

Native apps on a core bundle predating DEV-667 ignore `options.apiUrl` and degrade to same-region redemption rather than erroring.

### Author concerns

**1. The form branch has since been exercised — but region isolation still is not falsifiable.** A US-only user signing into an EU app completed sign-in through the form branch, and the **204 CORS preflight on the mint** proves `exchangeBase` was cross-origin: the auth page is served from the API host, so a same-origin mint does not preflight. Mint and redeem both targeted `us.api.fliplet.test`. What remains unproven is that a *wrong* host would fail — both regions share one Redis locally, so a misdirected redemption still resolves. Host targeting is confirmed; isolation needs two genuinely separate regional stacks or a staging pass.

**2. Two bugs found only because the app was served from the apps host.** In a published app `Fliplet.Env.get('apiUrl')` is the apps-host proxy while `primaryApiUrl` is the API host. `Fliplet.API.request` defaults to the former, so leaving `apiUrl` unset sent the exchange to `apps.fliplet.test`, not the canonical API host — every fallback path (rollout skew, tampered `authHost`, non-https) was affected. Now pinned to `safeAuthHost(authHost) || getApiOrigin()`. In Studio preview both values are identical and this would have stayed invisible.

**3. The allowlist initially rejected non-production regional hosts.** `us.api.fliplet.test` failed the `*.fliplet.com` test and fell back to the app's own host — passing locally only because the shared Redis resolves either way. That is the worst failure shape: correct in production, silently wrong off-production, masking cross-region bugs in exactly the environments used to test them. Now also accepts a single-label sibling of the app's own API host with a matching protocol, derived from the app's host so a crafted value cannot widen it, and **scoped off production** — on `fliplet.com` the regional-host rule is the whole allowlist, and a sibling rule there would readmit the CDN origins it exists to exclude. Deliberately **not** a registrable-domain ("last two labels") comparison — for an app on `api.acme.co.uk` that reduces to `co.uk` and would admit any `*.co.uk`.

**5. The allowlist needed two further narrowings after review.** `cdn.api.fliplet.com` and `cdn-staging.api.fliplet.com` are a *single label* under `api.fliplet.com` — structurally indistinguishable from a region — so they passed the regional-host rule and would have received the `Authorization` header; they are now explicitly excluded, and their two-label siblings (`us.cdn`, `ca.cdn`) were already out. Separately, every rule matched on `hostname` (port-free) while the function returned `origin` (port-bearing), so the port is now validated too. Both are covered by tests.

**4. Depends on an API-side core fix — and this is the real deploy gate.** This relies on Fliplet/fliplet-api#8465 correcting a `navigate.js` regex that silently dropped `authState` from `Fliplet.Navigate.query` and corrupted the `state` nonce. Without it the widget's `handleSameTabReturn()` bails at the first gate and sign-in never completes. Note this does **not** degrade to the legacy contract — it hard-breaks — and the asset's cache-buster is version-tied, so it does not change when the file does. The API PR carries the required mitigation (bump the `fliplet-core` asset version or republish) and a `config.disable_auth_exchange` kill switch that holds the exchange off until that lands.

[DEV-1633]: https://weboo.atlassian.net/browse/DEV-1633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/js/build.js
+++ b/js/build.js
@@ -490,8 +490,11 @@ Fliplet.Widget.instance('login', function(data) {
       xhr.onload = function() {
         var session;
 
+        var response;
+
         try {
-          session = JSON.parse(xhr.responseText).session;
+          response = JSON.parse(xhr.responseText);
+          session = response.session;
         } catch (err) {
           return fail('response was not JSON', xhr.status);
         }
@@ -500,16 +503,20 @@ Fliplet.Widget.instance('login', function(data) {
           return fail('unexpected response shape', xhr.status);
         }
 
-        // Pass the user through whole rather than hand-picking fields — the
-        // earlier six-field copy dropped data for deployed web/native sign-ins
-        // only. Note this is the session's user (models/session.js getPublic(),
-        // which omits auth_token); it is NOT identical to the profile the
-        // legacy token+user contract carries, which the login route enriches
-        // with host/region/organization/isFirstLogin/mustVerifyEmail and setup
-        // status. Public `login` hook consumers therefore still see a smaller
-        // object on this path — normalising the two shapes needs its own
-        // decision, since narrowing the legacy one would break consumers.
-        resolve({ token: session.auth_token, user: session.user });
+        // Prefer the enriched profile the API builds for a state redemption
+        // (host, region, organization, policy, setup status, isFirstLogin,
+        // mustVerifyEmail), so Fliplet.Hooks.run('login') gets the same shape
+        // the legacy token+user contract carried. It is built server-side from
+        // the authenticated user, not echoed back from the mint.
+        //
+        // Falls back to the session's user when absent — an API predating
+        // DEV-1633's profile support, or a profile build that failed and was
+        // logged server-side. Sign-in still completes; the hook payload is
+        // just the smaller object.
+        resolve({
+          token: session.auth_token,
+          user: response.profile || session.user
+        });
       };
 
       xhr.onerror = function() {
--- a/js/build.js
+++ b/js/build.js
@@ -416,12 +416,18 @@ Fliplet.Widget.instance('login', function(data) {
     // options.apiUrl replaces the base instead, which is the supported retarget
     // (added in DEV-667). Cores predating it ignore the option and fall back to
     // the app's own host — same-region behaviour, not a broken request.
-    var opts = { url: 'v1/session?state=' + encodeURIComponent(authState) };
-    var safeHost = safeAuthHost(authHost);
-
-    if (safeHost) {
-      opts.apiUrl = safeHost;
-    }
+    //
+    // apiUrl is set unconditionally. Leaving it off does NOT fall back to the
+    // canonical API host: Fliplet.API.request's default base is
+    // Fliplet.Env.get('apiUrl'), which inside a published web app is the
+    // apps-host-proxied URL (https://apps.fliplet.com/) rather than the API
+    // host — the same distinction getApiHost() exists to paper over. Pinning it
+    // to getApiOrigin() keeps the exchange on the host that serves these routes
+    // and matches where the login URL itself was sent.
+    var opts = {
+      url: 'v1/session?state=' + encodeURIComponent(authState),
+      apiUrl: safeAuthHost(authHost) || getApiOrigin()
+    };
 
     return Fliplet.API.request(opts).then(function(response) {
       var session = response && response.session;
--- a/js/build.js
+++ b/js/build.js
@@ -536,7 +536,13 @@ Fliplet.Widget.instance('login', function(data) {
   function handleSameTabReturn() {
     var q = readAuthReturnParams();
 
-    if (!q.token && !q.authState) return false;
+    // `error` counts as a return. A failed exchange comes back here with
+    // error+state and no credential; bailing out treated that as "not a return
+    // leg", so consumeAuthState() never ran and the URL was never cleaned. The
+    // nonce then survived in BOTH sessionStorage and the URL after being
+    // spent — and since the legacy token+user contract is still accepted, a
+    // leaked one could be replayed against a forged legacy callback.
+    if (!q.token && !q.authState && !q.error) return false;
 
     // Always consume the stored state, even on rejection — burning the
     // nonce on first arrival prevents replay if an attacker manages to
@@ -558,6 +564,18 @@ Fliplet.Widget.instance('login', function(data) {
 
     if (!expectedState || !q.state || q.state !== expectedState) {
       return reject('state nonce missing or mismatch');
+    }
+
+    // Checked after the nonce, so the nonce is consumed and the URL cleaned on
+    // this path too. No toast: the message is already rendered inline at mount
+    // (.login-error-holder), so a toast would just say the same thing twice.
+    if (q.error) {
+      // eslint-disable-next-line no-console -- security trace: failed returns need to surface for incident triage
+      console.warn('[Fliplet.Login] same-tab return carried an error:', q.error);
+      cleanAuthReturnParamsFromUrl();
+      showStart();
+
+      return true;
     }
 
     // Preferred contract: one-shot state exchange — no credentials in the URL.
--- a/js/build.js
+++ b/js/build.js
@@ -27,10 +27,16 @@ Fliplet.Widget.instance('login', function(data) {
     scheduleCheck();
   });
 
-  if (Fliplet.Navigate.query.error) {
+  // Read from location.search for the same reason handleSameTabReturn does —
+  // see readAuthReturnParams. A failed exchange returns here carrying ?error=,
+  // so this is part of the sign-in path and must not depend on which
+  // navigate.js the browser has cached.
+  var initialReturnError = readAuthReturnParams().error;
+
+  if (initialReturnError) {
     // .text(), never .html(): the error is a URL query param, so .html()
     // would be a reflected XSS sink on every app's login screen.
-    _this.$container.find('.login-error-holder').text(Fliplet.Navigate.query.error).addClass('show');
+    _this.$container.find('.login-error-holder').text(initialReturnError).addClass('show');
   }
 
   // ──────────────────────────────────────────────────────────────────────
@@ -141,6 +147,39 @@ Fliplet.Widget.instance('login', function(data) {
    * Strips the auth-return params from the URL in place so the token
    * doesn't persist in the address bar / history / bookmarks.
    */
+  // Reads the sign-in return params straight off window.location.search rather
+  // than via Fliplet.Navigate.query.
+  //
+  // Navigate.query is produced by fliplet-core's navigate.js, which until the
+  // paired API fix stripped internal params with an unanchored lookahead: it
+  // fired on the literal "&auth" inside ANY param starting with "auth", so
+  // "?state=<nonce>&authState=<uuid>" became "?state=<nonce>State=<uuid>" —
+  // authState dropped and the nonce corrupted. That asset carries a
+  // version-pinned cache-buster which does not change when the file does, so a
+  // browser holding the old copy would fail sign-in outright rather than
+  // degrading to the legacy contract.
+  //
+  // location.search is the raw browser value and no parser touches it, so
+  // reading it here makes the widget independent of which navigate.js the
+  // client happens to have cached. The API-side fix still matters for every
+  // other consumer of Navigate.query; this just removes the hard dependency
+  // from the sign-in path.
+  function readAuthReturnParams() {
+    var params = { token: null, user: null, state: null, authState: null, authHost: null, error: null };
+
+    try {
+      var qs = new URLSearchParams(window.location.search);
+
+      Object.keys(params).forEach(function(key) {
+        params[key] = qs.get(key);
+      });
+    } catch (err) {
+      console.warn('[Fliplet.Login] failed to parse return params:', err);
+    }
+
+    return params;
+  }
+
   function cleanAuthReturnParamsFromUrl() {
     if (!window.history || !window.history.replaceState) return;
 
@@ -390,74 +429,98 @@ Fliplet.Widget.instance('login', function(data) {
     }
 
     // Retarget via options.apiUrl — NOT by building an absolute options.url.
-    // Fliplet.API.request prepends the API base to *every* url, absolute ones
-    // included (core.js `options.url = apiUrl + options.url`), so an absolute
-    // URL yields https://api.fliplet.com/https://us-api.fliplet.com/... → 404.
-    // options.apiUrl replaces the base instead, which is the supported retarget
-    // (added in DEV-667). Cores predating it ignore the option and fall back to
-    // the app's own host — same-region behaviour, not a broken request.
+    // Issued as a plain XHR rather than through Fliplet.API.request, because
+    // that helper cannot express this call safely on every core version:
     //
-    // apiUrl is set unconditionally. Leaving it off does NOT fall back to the
-    // canonical API host: Fliplet.API.request's default base is
-    // Fliplet.Env.get('apiUrl'), which inside a published web app is the
-    // apps-host-proxied URL (https://apps.fliplet.com/) rather than the API
-    // host — the same distinction getApiHost() exists to paper over. Pinning it
-    // to getApiOrigin() keeps the exchange on the host that serves these routes
-    // and matches where the login URL itself was sent.
+    //  - An absolute options.url is prepended to the API base anyway
+    //    (core.js `options.url = apiUrl + options.url`, no scheme guard),
+    //    yielding https://api.fliplet.com/https://us.api.fliplet.com/… → 404.
+    //    That is the bug that got the first attempt at this feature reverted.
+    //  - The supported retarget, options.apiUrl, only landed in DEV-667
+    //    (Sep 2025). Cores predating it silently ignore it and send the
+    //    request to the app's OWN region, where a cross-region state does not
+    //    exist — so those users get a 401 and cannot sign in. Published native
+    //    bundles pin their core version, so that is not a hypothetical.
+    //
+    // Talking to the validated origin directly removes the version dependency
+    // and the double-prefix hazard together. The URL is fully controlled here:
+    // the origin has already passed safeAuthHost, and the state is encoded.
     var appOrigin = getApiOrigin();
+    var base = (safeAuthHost(authHost, appOrigin) || appOrigin).replace(/\/+$/, '');
+    var url = base + '/v1/session?state=' + encodeURIComponent(authState);
 
-    var opts = {
-      url: 'v1/session?state=' + encodeURIComponent(authState),
-      apiUrl: safeAuthHost(authHost, appOrigin) || appOrigin,
-      // The state token must be the ONLY credential this call presents.
-      //
-      // core.js fills in Fliplet.User.getAuthToken() whenever Auth-token is
-      // unset, and the API's loadUser keeps that ambient token when the state
-      // fails to resolve (expired, already consumed, Redis blip) instead of
-      // rejecting. The exchange would then return 200 with the device's
-      // PREVIOUS session and sign the user in as the wrong identity, silently
-      // — the exact opposite of what a failed exchange should do.
-      //
-      // Sending a deliberately invalid sentinel suppresses the fill-in, so an
-      // unresolvable state has nothing to fall back to and correctly 401s.
-      headers: { 'Auth-token': 'state' }
-    };
+    return new Promise(function(resolve) {
+      var xhr = new XMLHttpRequest();
 
-    return Fliplet.API.request(opts).then(function(response) {
-      var session = response && response.session;
-      var user = session && session.user;
+      function fail(reason, status) {
+        // Log the status and the host actually targeted. Without them CORS,
+        // network failure, an expired state and a wrong-region redemption all
+        // collapse into the caller's single "exchange failed" warn — and the
+        // cross-region path is the one that cannot be falsified on a local
+        // stack, so a production failure needs to be diagnosable from the log
+        // line alone.
+        console.warn('[Fliplet.Login] authState exchange failed', {
+          reason: reason,
+          status: status || 'none',
+          apiUrl: base,
+          authHostAccepted: !!safeAuthHost(authHost, appOrigin)
+        });
 
-      if (!session || !session.auth_token || !user) {
-        return null;
+        resolve(null);
       }
 
-      // Pass the user through whole rather than hand-picking fields. It is
-      // already the server's own public projection (models/session.js
-      // getPublic() omits auth_token), and it is the same object the legacy
-      // token+user contract puts in the callback URL — so every path feeds
-      // Fliplet.Hooks.run('login') an identically shaped userProfile.
-      // Reducing it here silently dropped fields for deployed web/native
-      // sign-ins only, which is exactly the kind of divergence a public hook
-      // must not have.
-      return {
-        token: session.auth_token,
-        user: user
-      };
-    }).catch(function(err) {
-      // Log the status and the host that was actually targeted. Without them
-      // CORS, network failure, an expired state and a wrong-region redemption
-      // all collapse into the caller's single "exchange failed" warn — and the
-      // cross-region path is precisely the one that cannot be falsified on a
-      // local stack, so a production failure here needs to be diagnosable from
-      // the log line alone.
-      console.warn('[Fliplet.Login] authState exchange failed', {
-        status: (err && (err.status || (err.response && err.response.status))) || 'none',
-        apiUrl: opts.apiUrl,
-        authHostAccepted: !!safeAuthHost(authHost, appOrigin),
-        message: (err && err.message) || String(err)
-      });
+      try {
+        xhr.open('GET', url, true);
+      } catch (err) {
+        return fail('open threw: ' + ((err && err.message) || err));
+      }
 
-      return null;
+      // The state token must be the ONLY credential this call presents.
+      //
+      // Without a header the API's loadUser falls back to the auth_token
+      // cookie when the state fails to resolve (expired, already consumed,
+      // Redis blip) — and on a same-origin exchange the browser sends that
+      // cookie regardless of withCredentials. The response would then be 200
+      // with the device's PREVIOUS session, signing the user in as the wrong
+      // identity. A deliberately invalid sentinel occupies req.auth_token so
+      // the cookie fallback never fires and an unresolvable state 401s.
+      xhr.setRequestHeader('Auth-token', 'state');
+      xhr.timeout = 15000;
+
+      xhr.onload = function() {
+        var session;
+
+        try {
+          session = JSON.parse(xhr.responseText).session;
+        } catch (err) {
+          return fail('response was not JSON', xhr.status);
+        }
+
+        if (xhr.status !== 200 || !session || !session.auth_token || !session.user) {
+          return fail('unexpected response shape', xhr.status);
+        }
+
+        // Pass the user through whole rather than hand-picking fields — the
+        // earlier six-field copy dropped data for deployed web/native sign-ins
+        // only. Note this is the session's user (models/session.js getPublic(),
+        // which omits auth_token); it is NOT identical to the profile the
+        // legacy token+user contract carries, which the login route enriches
+        // with host/region/organization/isFirstLogin/mustVerifyEmail and setup
+        // status. Public `login` hook consumers therefore still see a smaller
+        // object on this path — normalising the two shapes needs its own
+        // decision, since narrowing the legacy one would break consumers.
+        resolve({ token: session.auth_token, user: session.user });
+      };
+
+      xhr.onerror = function() {
+        fail('network or CORS failure', xhr.status);
+      };
+
+      xhr.ontimeout = function() {
+        fail('timed out after 15000ms');
+      };
+
+      xhr.send();
     });
   }
 
@@ -471,9 +534,9 @@ Fliplet.Widget.instance('login', function(data) {
    * normal session-restore path.
    */
   function handleSameTabReturn() {
-    var q = Fliplet.Navigate.query;
+    var q = readAuthReturnParams();
 
-    if (!q || (!q.token && !q.authState)) return false;
+    if (!q.token && !q.authState) return false;
 
     // Always consume the stored state, even on rejection — burning the
     // nonce on first arrival prevents replay if an attacker manages to
--- a/js/build.js
+++ b/js/build.js
@@ -617,8 +617,13 @@ Fliplet.Widget.instance('login', function(data) {
       var token = parsed.searchParams.get('token');
       var authState = parsed.searchParams.get('authState');
       var authHost = parsed.searchParams.get('authHost');
+      // A failed mint on the auth page returns to this same callback carrying
+      // only error+state. Without it in this guard the return is never marked
+      // handled: the IAB stays open on the callback page and the app sits
+      // behind it waiting for a result that never arrives.
+      var error = parsed.searchParams.get('error');
 
-      if (!token && !authState) return;
+      if (!token && !authState && !error) return;
 
       iabHandled = true;
 
@@ -643,6 +648,14 @@ Fliplet.Widget.instance('login', function(data) {
 
       if (!expectedState || !returnedState || returnedState !== expectedState) {
         return rejectIab('state nonce missing or mismatch');
+      }
+
+      // Checked after the nonce so a crafted ?error= link can't drive this
+      // path without a matching nonce. The server's message is logged rather
+      // than shown: it arrives via the URL, and the localised toast already
+      // says the same thing without rendering text from the query string.
+      if (error) {
+        return rejectIab('auth page reported an error: ' + error);
       }
 
       // Don't run the success path yet: on iOS a WebView navigation issued
--- a/js/build.js
+++ b/js/build.js
@@ -237,14 +237,50 @@ Fliplet.Widget.instance('login', function(data) {
       && /^token-[^@]*@fliplet\.com$/i.test(user.email);
   }
 
-  // Masks token / user / state / authState query params before logging
-  // the URL, so credentials don't surface in remote log aggregators,
-  // support-ticket screenshots, or screen recordings.
+  // Validates `authHost` before it is ever used as a request target.
+  //
+  // On the same-tab leg authHost arrives on the page query string, which the
+  // user (or anyone who can get them to open a crafted link) controls. The
+  // exchange sends the session's Authorization header to that host and feeds
+  // the response straight into handleAuthSuccess — which shape-validates the
+  // user but cannot validate its authenticity. An unvalidated host therefore
+  // buys an attacker both credential exfiltration and session fixation, so
+  // anything not provably Fliplet-owned is discarded and the caller falls
+  // back to the app's own API host (same-region behaviour).
+  //
+  // The app's own API origin is always accepted: it is the fallback target
+  // regardless, and on dev environments / local stacks it is not a
+  // fliplet.com host (e.g. https://api.fliplet.test).
+  function safeAuthHost(value) {
+    if (!value) return null;
+
+    try {
+      var u = new URL(value);
+
+      // Same origin as the app's own API host — trusted by definition.
+      if (u.origin === getApiOrigin()) return u.origin;
+
+      // Anything else must be HTTPS and Fliplet-owned. Matching on hostname
+      // (not the raw string) means "https://evil.com/?x=.fliplet.com" and
+      // "https://fliplet.com.evil.com" both fail.
+      if (u.protocol !== 'https:') return null;
+
+      return /(^|\.)fliplet\.com$/i.test(u.hostname) ? u.origin : null;
+    } catch (err) {
+      return null;
+    }
+  }
+
+  // Masks token / user / state / authState / authHost query params before
+  // logging the URL, so credentials don't surface in remote log aggregators,
+  // support-ticket screenshots, or screen recordings. authHost is not itself
+  // a credential, but it is masked alongside them so a log line can't be used
+  // to confirm which region a given user's session lives in.
   function maskUrlForLogging(url) {
     try {
       var u = new URL(url);
 
-      ['token', 'user', 'state', 'authState'].forEach(function(key) {
+      ['token', 'user', 'state', 'authState', 'authHost'].forEach(function(key) {
         if (u.searchParams.has(key)) u.searchParams.set(key, '<redacted>');
       });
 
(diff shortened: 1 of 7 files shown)


Arpanexe:
Review findings for `0fac779a`:

1. **[P1] Cached `navigate.js` can break the same-tab callback.**

   `handleSameTabReturn()` relies on `Fliplet.Navigate.query`, but the parser fix exists only in paired API PR #8465, whose cache key reportedly does not change. The old parser corrupts `?state=n&authState=s` into `?state=nState=s`, so `authState` disappears and the widget skips the exchange. Existing web users can therefore remain unable to sign in until a hard refresh or app republish.

   Please parse these callback parameters directly from `window.location.search`, or ship a reliable cache-busting/version change before enabling the exchange.

2. **[P2] Cross-region native login still fails on older core bundles.**

   `exchangeAuthState()` depends on `options.apiUrl`. Older native cores ignore that option, sending the request to the app's region even though the Redis state exists only in the issuer's region. The error is swallowed and becomes a generic login failure.

   Please use a direct request to the validated origin, or feature-detect support and retain the legacy contract where unsupported.

3. **[P2] The exchange changes the public login-hook payload.**

   The exchanged user is reduced to six fields, while popup and legacy paths pass the full auth-loader profile into `Fliplet.Hooks.run('login')`. Fields such as region, organization, policy, setup/first-login data, and `impersonatedFrom` disappear only on deployed web/native flows. The bundled listener in `js/app.js` already reads `userProfile.impersonatedFrom`.

   Please normalize all paths to the same public profile before firing the hook.

Verification: the PR head passes `node --check`. The widget has no automated coverage for this flow, and Codacy currently reports `ACTION_REQUIRED`.

js/build.js
```diff
@@ -346,17 +412,82 @@ Fliplet.Widget.instance('login', function(data) {
...
+    // apiUrl is set unconditionally. Leaving it off does NOT fall back to the
+    // canonical API host: Fliplet.API.request's default base is
+    // Fliplet.Env.get('apiUrl'), which inside a published web app is the
+    // apps-host-proxied URL (https://apps.fliplet.com/) rather than the API
+    // host — the same distinction getApiHost() exists to paper over. Pinning it
+    // to getApiOrigin() keeps the exchange on the host that serves these routes
+    // and matches where the login URL itself was sent.
+    var opts = {
+      url: 'v1/session?state=' + encodeURIComponent(authState),
+      apiUrl: safeAuthHost(authHost) || getApiOrigin()
```
hcarneiro: **Critical — an expired or already-consumed `authState` silently resolves to the device's previous session instead of failing.**

`libs/authenticate.js:559-562` in fliplet-api:

```js
const storedToken = await database.redis.evalAsync(luaScript, 1, `sso:state:${state}`);

if (storedToken) {
  req.auth_token = storedToken;
}
```

No `else`. When the state doesn't resolve, `req.auth_token` keeps whatever the header supplied — and `Fliplet.API.request` **always** supplies one (`core.js:1789-1791`):

```js
if (!options.headers['Auth-token']) {
  options.headers['Auth-token'] = Fliplet.User.getAuthToken();
}
```

So `GET /v1/session?state=<expired>` returns **200 with the device's cached session**, not 401. `exchangeAuthState` checks only that `session.auth_token` and `session.user` exist — both do — and `handleAuthSuccess` runs.

**Failure scenario:** user A is signed in and clicks sign-in to switch to user B. The exchange expires (>60s on the auth page, a Redis blip, or the 5s grace elapsing after a double-request). The widget reports success and signs them in **as A**, silently, with no error anywhere. This didn't exist under the legacy contract, where the identity came from the URL and a missing token simply failed.

On native IAB the shared cookie jar makes the fallback token the app bootstrap token — that specific case is caught by the existing `isAppTokenUser` guard, but a real prior user session is not.

Two fixes, pick one:

**Preferred, API-side** — in `loadUser`, when `req.query.state` is present but unresolvable, clear `req.auth_token` and let the request 401. Correct semantics (a caller presenting `?state=` is asserting that credential), but `authenticate.js` is a hotspot and this changes behaviour for every `?state=` consumer (preview, downloads, V3 attachments), so it wants its own regression pass.

**Scoped to this PR** — stop `core.js` attaching the ambient token on this call, so an unresolved state genuinely 401s:

```js
    var opts = {
      url: 'v1/session?state=' + encodeURIComponent(authState),
      apiUrl: safeAuthHost(authHost) || getApiOrigin(),
      // The state token is the only credential this call may present.
      // core.js fills in Fliplet.User.getAuthToken() when Auth-token is
      // unset, and authenticate falls back to it when the state fails to
      // resolve — which would complete sign-in as the device's previous
      // identity instead of failing.
      headers: { 'Auth-token': 'state' }
    };
```

Either way, add a test for "expired state → sign-in fails" — it's the whole failure mode.

js/build.js
```diff
@@ -229,14 +237,72 @@ Fliplet.Widget.instance('login', function(data) {
...
+      var app = new URL(appOrigin);
+
+      // Same origin as the app's own API host — trusted by definition.
+      if (u.origin === appOrigin) return u.origin;
+
+      // Production regional hosts (us-api.fliplet.com, ca-api.fliplet.com…).
+      // Matching on the parsed hostname (not the raw string) means
+      // "https://evil.com/?x=.fliplet.com" and "https://fliplet.com.evil.com"
+      // both fail.
+      if (u.protocol === 'https:' && /(^|\.)fliplet\.com$/i.test(u.hostname)) {
```
hcarneiro: **Warning (possible, not confirmed) — this accepts *any* `https://*.fliplet.com`.**

This is what the ticket prescribed and the parsing is done correctly. But the set of legitimate targets is small and known — the regional API hosts — while `*.fliplet.com` covers a lot of surface: `apps.fliplet.com`, and the 13 `*.studio-apps.fliplet.com` origins already special-cased in `libs/cors.js:15-28`. This call carries `Authorization: Bearer` and feeds the response into `handleAuthSuccess`.

I have **not** confirmed a customer-controlled `*.fliplet.com` origin exists, so I'm marking this possible rather than a defect. Tightening costs nothing and removes the question:

```suggestion
      if (u.protocol === 'https:' && /^([a-z0-9-]+-)?api\.fliplet\.com$/i.test(u.hostname)) {
```

The off-production sibling branch below (:288-292) already handles dev/local hosts, so this doesn't reintroduce the problem you fixed in concern #3.

---

**Also — no committed test for this function.** Fair point on not adopting a runner in this PR, but `safeAuthHost` was one of the two BLOCKING findings that caused the original revert, and the 18 hostile inputs you verified against live in a harness that isn't in the repo — so the next person to touch it has nothing.

Node has a test runner built in; no dependency, no config. Export the function (`if (typeof module !== 'undefined') module.exports = { safeAuthHost };` at the bottom of the IIFE, or split it into `js/safe-auth-host.js`) and commit the harness you already wrote as `test/safe-auth-host.test.js` run by `node --test`. That satisfies the AC's "widget-side coverage for the host handling if practical" without introducing a framework.

js/build.js
```diff
@@ -380,6 +511,28 @@ Fliplet.Widget.instance('login', function(data) {
...
     }
 
+    // Preferred contract: one-shot state exchange — no credentials in the URL.
+    if (q.authState) {
+      var authState = q.authState;
+      var authHost = q.authHost;
+
+      cleanAuthReturnParamsFromUrl();
+
+      exchangeAuthState(authState, authHost).then(function(result) {
```
hcarneiro, line 628: **Suggestion — no `.catch` on this chain.**

`exchangeAuthState` catches its own request failure, but a throw inside `handleAuthSuccess` becomes an unhandled rejection with no toast and no `showStart()`, leaving the loader up indefinitely. One line closes it:

```js
      }).catch(function() {
        reject('exchange threw');
      });
```

hcarneiro (changes requested):
## Code review (via `/fliplet-pr-review`)

Reviewed together with Fliplet/fliplet-api#8465 — neither works alone.

### Summary

Both blocking findings from the reverted DEV-1572 attempt are genuinely fixed. The `options.apiUrl` retarget is the right call — `core.js:1784` does `options.url = apiUrl + options.url` with no scheme guard, so the old absolute-URL approach really did 404. `safeAuthHost()` is carefully built: matching on `u.hostname` rather than the raw string means `https://fliplet.com.evil.com` and `https://evil.com/?x=.fliplet.com` both fail, and deliberately not reducing to a registrable domain is the correct call for apps on a public suffix.

One Critical below, plus the API-side Critical this PR depends on.

### Regression Risk

Moderate and contained to the login widget's return leg, but it is the auth path: a wrong outcome is a wrong signed-in identity, not a rendering glitch. Native IAB is the higher-risk half — the shared cookie jar means the ambient token in play is the app bootstrap token.

### Critical — an expired or already-consumed `authState` silently signs the user in as the device's *previous* identity

Inline on `js/build.js`. `libs/authenticate.js:559-562` (fliplet-api) has no `else` when the Redis lookup misses, so `req.auth_token` keeps the header token that `Fliplet.API.request` always attaches (`core.js:1789-1791`). The exchange therefore returns **200 with the cached session**, not 401, and `exchangeAuthState` can't tell the difference.

### Blocked on Fliplet/fliplet-api#8465

That PR patches only one of the two `/v1/auth/login` redirect builders — `libs/oauthHelpers.js:945` (google/microsoft/apple) still drops `authExchange`, so social SSO through this widget falls back to `token=…&user=…` in the callback URL. Nothing to change here; noting it because the AC "return URL carries `authState` only" is not met until it lands.

### Acceptance Criteria (DEV-1633)

| AC | Verdict |
|---|---|
| Return URL carries `authState` only, never `token`/`user` | ❌ Not met for Google/Microsoft/Apple SSO (API-side). Met for password + SAML2 |
| Same-region AND cross-region login works | ⚠️ Partial — the form branch (`result.user.host` after a 2FA region rebind) is unexercised, as you note |
| `authHost` validated against an allowlist | ✅ `js/build.js:254-300` |
| CORS confirmed for cross-region redemption | ✅ `libs/config.js:159` concats every `region.host`; `libs/cors.js` sets no `allowedHeaders` so request headers are reflected. Matches your live 204 |
| Automated coverage where feasible | ❌ None — see inline on `safeAuthHost` |

### Notes

- Your concerns #2 and #3 are both good catches that only surfaced because you tested off Studio preview. #3 in particular (allowlist silently wrong off-production, correct in prod) is the worst possible failure shape and you were right to fix it rather than ship it.
- Concern #1 is real: a local stack cannot falsify the form branch because both regions share one Redis, so a wrong-host redemption still resolves. Not a blocker, but it should not be closed as "verified" — it needs two genuinely isolated regional stacks or a staging pass.
- Branch strategy is fine. Widgets normally PR direct to `master`, but this repo already ran the integration-branch flow for DEV-1572 (`projects/DEV-1572` → `release/DEV-1572`), so `projects/DEV-1633` is consistent. Just don't forget the `release/DEV-1633` → `master` PR afterwards.

### CI

**Red** — `Codacy Static Code Analysis` reports `action_required` on `0fac779a`. Looks like an authorization/re-run state rather than a genuine finding, but it needs to go green or be confirmed as unrelated infra before merge.

### Verdict

**REQUEST_CHANGES** — the Critical below is a silent wrong-identity outcome introduced by moving from URL-delivered credentials to a server-side resolution that fails open, and `safeAuthHost` (one of the two findings that caused the original revert) ships with no committed test.

zeryabkhan91:
Thanks both. Everything below is on `151210d`. The allowlist finding was the important one and it turned out to be worse than either of us described.

## @hcarneiro

**Critical — expired/consumed `authState` signs the user in as the previous identity.** Confirmed exactly as you traced it. Took the scoped fix: `js/build.js:414` now sends `Auth-token: 'state'`, so `core.js` doesn't fill in the ambient token and an unresolvable state has nothing to fall back to. Reasoning for declining the API-side fix is on Fliplet/fliplet-api#8465 — short version, `?state=` has five other consumers and that change wants its own regression pass. Follow-up ticket coming.

**Warning — the allowlist accepts any `https://*.fliplet.com`.** Fixed, but **not with the suggested pattern** — `/^([a-z0-9-]+-)?api\.fliplet\.com$/i` would have broken cross-region in production. The real regional hosts are dot-separated:

```
us  https://us.api.fliplet.com/
ca  https://ca.api.fliplet.com/
```

That regex requires a hyphen before `api`, so it rejects both, and they'd fall back silently to the EU host — reintroducing exactly the bug this PR exists to fix, in the failure shape you correctly called the worst one. Took the intent with `\.` instead: `/^(?!cdn[.-])([a-z0-9-]+\.)?api\.fliplet\.com$/i`, which covers all four production hosts plus staging.

The `(?!cdn[.-])` is there because the narrowing wasn't enough on its own. `cdn.api.fliplet.com` and `cdn-staging.api.fliplet.com` are a **single label** under `api.fliplet.com` — structurally indistinguishable from a region — so they passed the narrowed rule and would have received the `Authorization` header. All four CDN origins are real hosts in `config/production.json`. Their two-label siblings (`us.cdn`, `ca.cdn`) were already excluded.

So your instinct that the legitimate target set is small and known was right, and it needed a second guard to actually get there.

**No committed test for `safeAuthHost`.** Fixed, and I took your suggestion of the built-in runner. Extracted to `js/safe-auth-host.js` taking the app origin as an argument — which makes it pure and also removes a double `getApiOrigin()` call — with `test/safe-auth-host.test.js` run by `node --test` via `npm test`. No new dependency. **18 tests**, covering the hostile inputs that previously only lived in a throwaway harness plus everything both review rounds surfaced: the CDN origins, subdomain depth, ports, and trailing-dot FQDNs.

**Suggestion — no `.catch` on the exchange chain.** Fixed in both places. The same-tab path first, then `runPendingAuth` on the native IAB path, which had the identical shape and the same consequence.

**Your note on `exchangeAuthState` swallowing the status and host.** Fixed — it now logs `status`, `apiUrl`, `authHostAccepted` and `message`, so CORS, network failure, expired state and wrong-region are distinguishable. Agreed this matters most on the branch that can't be falsified locally.

**Your note on the branch strategy** — agreed, `release/DEV-1633` → `master` to follow.

## @Arpanexe

**1. Cached `navigate.js` can break the same-tab callback.** Confirmed, and this is the real deploy gate — it hard-breaks rather than degrading, exactly as you said. Not fixable from this repo, so it's handled on the API PR: a **required** mitigation (bump the `fliplet-core` asset version or republish) rather than an advisory note, plus a server-side `config.disable_auth_exchange` kill switch that holds the exchange off until that lands.

**2. Older native cores ignore `options.apiUrl`.** Still open, and I want to be straight about it rather than claim otherwise. I haven't added feature detection. `options.apiUrl` predates this work (DEV-667), but I have no floor on what core version published native apps actually carry, so I can't currently say how much exposure there is. The kill switch means this is recoverable without a republish if it bites, but that's mitigation, not a fix. Happy to add a capability check if you think the risk is real — do you have a view on the oldest core still in the wild?

**3. The exchange changes the public login-hook payload.** Fixed — the user now passes through whole rather than being reduced to six fields, so every path feeds `Hooks.run('login')` an identically shaped `userProfile`. It's already the server's own public projection (`models/session.js` `getPublic()` omits `auth_token`) and the same object the legacy contract puts in the URL.

One correction on the specific example: `impersonatedFrom` doesn't appear in any API response — it's set by `public/assets/fliplet-native/1.0/js/authentication.js` onto `window.ENV.user`, a different object — so `js/app.js:344` was reading `undefined` on the popup path too. The general point about divergence stood and is what I fixed.

## CI

Codacy is red with `conclusion=action_required`, zero annotations and no output summary — and it's identical on **every** open PR in this repo (#99 through #106). That reads as an integration/authorization state rather than a finding against this branch. Needs someone with Codacy access to reconnect it or an explicit waiver; I can't clear it from here.

There's also no CI running the new tests — `npm test` passes locally but nothing enforces it on push. Happy to add a workflow, though it felt like a separate change from this PR.

Arpanexe:
Second review round for `151210d` (reviewed directly, without the Fliplet PR review skill):

1. **[P1] Still open — cached `Navigate.query` breaks same-tab login.**

   `handleSameTabReturn()` still reads `Fliplet.Navigate.query`. Browsers retaining the old `navigate.js` parser lose `authState` and corrupt the nonce (`?state=n&authState=s` becomes `?state=nState=s`), so this function skips the exchange. The paired API fix has a cache key that reportedly does not change, leaving existing web users dependent on a hard refresh or app republish.

   Please parse `window.location.search` directly here, or guarantee cache invalidation before enabling the exchange.

2. **[P2] Still open — older native cores cannot redeem cross-region states.**

   The request still relies on `options.apiUrl`, introduced by DEV-667 in September 2025. Older native bundles ignore it, target the app's region, and receive 401 because the state resides only in the issuer region. The new sentinel header makes this fail closed, but it still prevents the user from signing in.

   A direct request to the validated origin would avoid the core-version dependency.

3. **[P2] The hook-payload fix is incomplete.**

   Passing `session.user` through whole is an improvement, but it does not make the exchange payload identical to the legacy auth-loader `result.user`, despite the new comment. The legacy `/v1/auth/login` response additionally contains fields such as `host`, `region`, `organization`, `isFirstLogin`, `mustVerifyEmail`, and setup status; those are not properties of the session's associated user. Public `login` hooks therefore still receive different profiles depending on the return flow.

   Please return the legacy profile shape from the exchange or explicitly normalize every flow to one documented shape before firing the hook.

4. **[P2] New — the committed test command fails.**

   `package.json` defines `"test": "node --test test/"`. On Node 22 this treats `test/` as a module path and fails with:

   ```
   Error: Cannot find module '<repo>\\test'
   ```

   `node --test` and `node --test test/safe-auth-host.test.js` both pass all 18 tests. The new test file also cannot be parsed by the repository's ES5 ESLint configuration because it starts with `const`/destructuring; this may be related to Codacy remaining `ACTION_REQUIRED`.

   Please change the script to `node --test` (or the explicit filename) and either add a test-file ESLint override or write the test in the configured syntax level.

Positive verification: the ambient-token/wrong-identity vulnerability is fixed; the host validation now has meaningful coverage; direct execution passes 18/18; all three changed JS files pass `node --check`. Codacy remains `ACTION_REQUIRED` on this head.

zeryabkhan91:
@Arpanexe — all four addressed on `8b3dd9f`. You were right to keep pushing on 1 and 2; both had a real fix and I'd been treating them as deploy-process problems.

**1. Cached `Navigate.query` breaks same-tab login — fixed properly.** Took your suggestion. `handleSameTabReturn` now reads `window.location.search` via a local `readAuthReturnParams()` instead of `Fliplet.Navigate.query`. `location.search` is the raw browser value and no parser touches it, so the sign-in path is independent of which `navigate.js` the client has cached. Applied to the initial `?error=` render at the top of the widget too, since a failed exchange returns through it.

This removes the hard-break combination entirely rather than relying on the cache-invalidation step in the API description. That step still matters for every other `Navigate.query` consumer, so it stays.

**2. Older native cores can't redeem cross-region states — fixed, but not the way suggested.** A direct request *through `Fliplet.API.request`* is the trap: `core.js` prepends the API base to an absolute `options.url` with no scheme guard, giving `https://api.fliplet.com/https://us.api.fliplet.com/…` → 404. That is precisely what got the first attempt at this feature reverted.

So the exchange is now a plain `XMLHttpRequest` to the validated origin, bypassing `core.js` entirely. That removes the `options.apiUrl` version dependency and the double-prefix hazard together, and lets me set the headers exactly.

One thing worth recording: the `Auth-token` sentinel is still required and I nearly dropped it. On a **same-origin** exchange the browser sends the `auth_token` cookie regardless of `withCredentials`, so without the sentinel an unresolvable state would still fall back to the device's previous session — the original wrong-identity bug, reintroduced by a change meant to be unrelated.

**3. Hook payload fix is incomplete — you're right, and my comment was wrong.** I verified: `routes/v1/auth.js:802-815` builds the legacy profile from `_.pick(user, [...])` and then adds `host`, `region`, `trusted`, `organization`, `isFirstLogin`, `mustVerifyEmail` plus `addUserSetupStatus()`. None of those are properties of the session's associated user, so passing `session.user` through whole is closer but not equal, and claiming parity in the comment was worse than the original six-field copy — at least that didn't assert something false.

Comment corrected to state what the payload actually is and that the shapes still differ. I've **not** normalised them here: making the exchange return the legacy shape means changing `GET /v1/session` for all its consumers, and narrowing the legacy path to match would break `login` hook consumers relying on those fields. That's a decision rather than a fix — happy to take it either way, but it wants to be deliberate and probably documented as the one public shape.

**4. `npm test` fails on Node 22 — fixed.** Changed to `node --test`, which works on both (verified 18/18 on Node 20 here, and you've confirmed 22). Also added an ESLint override so `test/**` parses at ES2022 without relaxing the ES5 level the shipped browser assets target — `js/` is unaffected.

On the Codacy link: I don't think the test file is the cause. `action_required` with zero annotations is present on **every** open PR in this repo (#99–#106), including several that predate this branch and this file. That reads as an integration/authorization state rather than a finding. Worth someone with Codacy access confirming, but it shouldn't block on the test file.

Note there's still no CI running these tests — `npm test` passes locally but nothing enforces it on push.

Arpanexe:
Third review round for `8b3dd9f` (reviewed directly, without the Fliplet PR review skill):

The cached-parser issue, old-core `options.apiUrl` dependency, and broken `npm test` command are fixed. Three findings remain:

1. **[P1] Error-only same-tab returns do not consume the nonce.**

   The API now returns failed exchanges with `error+state`, but `handleSameTabReturn()` still gates only on `token/authState`. The error is rendered at widget startup, then the handler returns false and falls into normal session restoration without clearing the URL or calling `consumeAuthState()`.

   This leaves the supposedly single-use nonce in both session storage and the URL. Because the legacy `token+user` callback remains accepted, a leaked nonce can still be reused to pass the state check on a forged legacy callback.

   Please include `q.error` in the handled-return guard, validate and consume its nonce, clean the URL, show the form, and return handled.

2. **[P2] The new ESLint configuration is invalid.**

   The override in `.eslintrc` contains a `comment` property, which ESLint's configuration schema does not permit. A focused lint run exits before inspecting any source file:

   ```text
   Error: ESLint configuration in .eslintrc is invalid:
     - Unexpected top-level property "overrides[0].comment".
   ```

   Please remove that property or convert the configuration to JavaScript and use an actual source comment.

3. **[P2] Login-hook payload compatibility remains unresolved.**

   The updated comment now accurately documents that `session.user` is smaller than the legacy auth-loader profile, but the runtime divergence remains. Public `login` hooks still lose legacy fields such as host, region, organization, first-login/email-verification data, and setup status on the exchanged web/native paths.

   This needs either normalization to one public profile or an explicit compatibility decision before release; correcting the comment alone does not preserve existing hook behavior.

Verification on the detached PR head: `npm test` passes 18/18, all changed JavaScript passes `node --check`, and `git diff --check` passes. ESLint cannot start because of finding #2. Codacy remains `ACTION_REQUIRED`.

zeryabkhan91:
@Arpanexe — all three fixed. Head is `77c61ef`, 18/18 pass.

## 1. Error-only same-tab returns did not consume the nonce (`97bcc56`)

Correct, and the sharpest finding of the round — the replay path is the part I would have missed. `handleSameTabReturn` gated on `token`/`authState`, so a failed exchange arriving with `error`+`state` was not treated as a return leg at all: `consumeAuthState()` never ran and the URL was never cleaned, leaving a *spent* single-use nonce in both sessionStorage and the address bar. With the legacy `token`+`user` contract still accepted, that is replayable against a forged callback.

`error` now counts as a handled return. Order matters and matches your description: consume the nonce, validate it, *then* branch on the error — so a crafted `?error=` link without a matching nonce still goes down the normal rejection path rather than getting a free pass. URL cleaned, form revealed, `true` returned. No toast on that branch, because the message is already rendered inline at mount via `.login-error-holder`; a toast would just say the same thing twice.

## 2. The ESLint override was invalid (`97bcc56`)

My mistake, and a bad one — I put a `comment` property in the override, which is not in the config schema, so ESLint refused to start *for the whole repo*. An override intended to let `test/` parse at ES2022 instead disabled linting entirely. Removed; the override is now `files` / `parserOptions` / `env` only.

Being straight about the cause: this repo has no `node_modules`, so I cannot run ESLint here and shipped config I had not validated. I should have said that rather than adding it silently — it is the second time this thread I have reported something about widget linting that did not hold up. Please treat any lint claim from me on this PR as unverified unless CI says otherwise.

## 3. Login-hook payload (`77c61ef` + Fliplet/fliplet-api#8465 `a76d1b75`)

Normalised rather than documented. The API now returns an enriched `profile` on a state redemption and the widget prefers it over `session.user`, falling back when absent so an older API or a failed profile build still completes sign-in with the smaller object.

Full reasoning is on the API PR, but two points belong here:

- The profile is **built server-side from the authenticated user**, not echoed back from the mint. Carrying it through the mint would have meant accepting `isAdmin` / `organization` / `policy` from anyone holding a valid `Auth-token` and feeding that straight into every app's public `login` hook. Only `isFirstLogin` and `trusted` ride along, because they cannot be recomputed after the login request.
- **The legacy contract was never one shape.** `routes/v1/auth.js:1667` hands the auto-success branch a five-field user; only the form branch carries the enriched profile. So there was no single legacy shape to match — this makes both exchange branches consistent and richer than legacy auto-success.

## Still open, neither in my hands

**No CI runs these tests.** You have raised it twice and you are right. `npm test` passes locally but nothing enforces it on push, so the 18 tests are only as good as whoever remembers to run them. Happy to add a workflow if that is wanted in this PR rather than a follow-up.

**Codacy** remains `ACTION_REQUIRED`. I do not think the test file is the cause: the same conclusion with zero annotations is on every open PR in this repo (#99–#106), including several that predate this branch. It needs someone with Codacy access to reconnect it or waive it explicitly.

PR title updated — it still claimed `options.apiUrl`, which this PR removed.

js/build.js
```diff
@@ -346,17 +439,147 @@ Fliplet.Widget.instance('login', function(data) {
...
+    var q = readAuthReturnParams();
 
-    if (!q || !q.token) return false;
+    // `error` counts as a return. A failed exchange comes back here with
+    // error+state and no credential; bailing out treated that as "not a return
+    // leg", so consumeAuthState() never ran and the URL was never cleaned. The
+    // nonce then survived in BOTH sessionStorage and the URL after being
+    // spent — and since the legacy token+user contract is still accepted, a
+    // leaked one could be replayed against a forged legacy callback.
+    if (!q.token && !q.authState && !q.error) return false;
```
hcarneiro, line 582: **Warning — this makes any `?error=` on the URL hijack the login screen.**

Admitting `error` into the handled-return guard is the right fix for the replay hole @Arpanexe found. But `error` is a generic param name and this gate has no other qualifier, so *any* page load carrying `?error=` — a deep link, another widget's param, an app's own redirect — is now treated as a sign-in return leg.

What follows on such a load:

1. `consumeAuthState()` burns the stored nonce (line 587, before the check)
2. the nonce check fails, so `reject()` fires: `Fliplet.UI.Toast.error(unableLogin)` + `showStart()` + URL cleaned
3. `handleSameTabReturn()` returns `true`, so `init()` returns at :1349 and **`initSession()` never runs** (:1353)

So a signed-in user opening a login-screen URL with an unrelated `?error=` gets a login-failure toast and is held on the sign-in form instead of being restored and redirected onward. Before this PR, `?error=` was rendered into `.login-error-holder` and execution continued normally — the param didn't touch the return-leg path at all.

Step 1 also stands on its own: a second tab loading such a URL burns the nonce a real in-flight sign-in is waiting on, so that sign-in then fails its nonce check.

The API always sends `error` together with `state` — `failExchange` redirects to `cbBase`, which already carries the nonce the widget appended — so requiring both is exact and costs one line:

```suggestion
    if (!q.token && !q.authState && !(q.error && q.state)) return false;
```

A genuine failed exchange still takes the path; an incidental `?error=` falls through to `initSession()` as before.

The mount-time render at :36 is fine as-is — swapping the server's string for the localised one is a good call, and the social-engineering rationale in that comment is the right reason for it.

hcarneiro (commented):
## Re-review (via `/fliplet-pr-review`) — head `0519c132`

Reviewed against the current head, not the thread. Every finding from my last pass is fixed, and I verified each rather than taking the replies on trust — including running the test suite: **18/18 pass** on the PR head here.

### Previous findings — all resolved

| Finding | Status |
|---|---|
| **Critical** — expired state signs the user in as the previous identity | ✅ Fixed. Verified the mechanism: `loadUser` reads the header first, the `'state'` sentinel occupies `req.auth_token`, and the cookie fallback is guarded on `!req.auth_token` — so an unresolvable state has nothing to fall back to. Your note that the same-origin cookie is sent regardless of `withCredentials`, so the sentinel is still required after the move to plain XHR, is correct and worth having in the comment |
| **Warning** — allowlist accepted any `https://*.fliplet.com` | ✅ Fixed — **and my suggested pattern was wrong.** `/^([a-z0-9-]+-)?api\.fliplet\.com$/i` requires a hyphen before `api`, so it would have rejected `us.api.fliplet.com` and `ca.api.fliplet.com` and fallen back silently to the EU host — the exact failure shape the PR exists to fix. Good catch, and the `(?!cdn[.-])` exclusion is a real hole I hadn't seen: `cdn.api.fliplet.com` is a single label under `api.fliplet.com` and structurally indistinguishable from a region |
| **Warning** — no committed test for `safeAuthHost` | ✅ Extracted to `js/safe-auth-host.js` as a pure function taking `appOrigin`, 18 tests under `node --test`, no new dependency, and a CI workflow so it can't rot. This was the right shape |
| **Warning** — CI red | ⚠️ The new `test` check is green. Codacy is still `action_required` — see below |
| **Suggestion** — no `.catch` on the exchange chain | ✅ Fixed in both places, and the native one does the fuller cleanup (toast + `hideLoadingState` + `showStart`) |

The port guard in `safeAuthHost` is a good addition nobody asked for — every rule matches on `hostname` while the function returns `origin`, so without it a host that satisfies a rule could still redirect the redemption to an arbitrary port. Worth the four lines.

The move to plain XHR is also the right resolution of the old-core problem: it removes the `options.apiUrl` version dependency and the absolute-URL double-prefix hazard at the same time, and the double-prefix trap is exactly what got the first attempt reverted.

### Regression Risk

Lower than last round on the auth path — the wrong-identity failure mode is closed and the host allowlist now has real coverage. It went **up** slightly on widget mount: `?error=` handling now changes control flow rather than just rendering text. That's the one inline finding below.

### New findings this round

One Warning, inline. No Criticals.

### CI

**Amber.** `test` passes (18/18, reproduced locally). `Codacy Static Code Analysis` is still `action_required` — it has reported no findings, no output title and no summary across five consecutive heads, which reads as an authorization/re-run state rather than a code failure. It needs a maintainer to re-run or re-authorize the integration; it isn't something to fix in the diff. Worth confirming that before merge rather than merging past a permanently amber check.

### Verdict

**COMMENT.** Nothing here blocks merge. The one Warning is a one-line fix and I'd take it, but it's not worth holding the PR for if you'd rather land the pair and follow up.

One process note: `re: your `.eslintrc` message` — flagging your own unverified lint claims was the right call, and the CI workflow you added means neither of us has to take a lint claim on trust again.

Arpanexe:
Re-review of latest head `d8862d3`:

**[P2] `.eslintrc` remains incompatible with the repository ESLint version**

[Line 12](https://github.com/Fliplet/fliplet-widget-fliplet-login/blob/d8862d3b8c6b496a98bee1485cdb8507640285c1/.eslintrc#L12) enables the `es2022` environment, but the lockfile resolves ESLint 8.2.0, which rejects it. Running `npx eslint js/safe-auth-host.js test/safe-auth-host.test.js` exits before linting with:

```
Environment key "es2022" is unknown
```

Please use the supported `es2021` environment while retaining `parserOptions.ecmaVersion: 2022`, or upgrade the locked ESLint version. The new workflow only runs tests and syntax checks, so its green result does not catch this.

Everything else reviewed cleanly, including the error-return guard and enriched profile handling.

Verification: 18/18 tests passed; JavaScript syntax checks and `git diff --check` passed.

js/build.js
```diff
@@ -27,10 +27,24 @@ Fliplet.Widget.instance('login', function(data) {
...
 
-  if (Fliplet.Navigate.query.error) {
-    // .text(), never .html(): the error is a URL query param, so .html()
-    // would be a reflected XSS sink on every app's login screen.
-    _this.$container.find('.login-error-holder').text(Fliplet.Navigate.query.error).addClass('show');
+  // Read from location.search for the same reason handleSameTabReturn does —
+  // see readAuthReturnParams. A failed exchange returns here carrying ?error=,
+  // so this is part of the sign-in path and must not depend on which
+  // navigate.js the browser has cached.
+  var initialReturnError = readAuthReturnParams().error;
```
hcarneiro: **Suggestion — the guard below learned that `error` alone is not a sign-in signal. This did not.**

After `d8862d3b8` a bare `?error=` no longer burns the nonce, no longer toasts, and no longer skips `initSession()` — all the damage is gone. But it still paints "Sign-in could not be completed. Please try again." into `.login-error-holder` on a page load that had nothing to do with signing in.

To be clear which part I mean: showing the localised string instead of the server's is right, and the reasoning in the comment below holds — I said so last round and still think it. This is about *when* it fires, not *what* it shows.

A real failed exchange always carries the nonce on `cbBase`, so gating this the same way the guard is now gated costs one line and changes nothing on the path that matters:

```suggestion
  var initialReturnParams = readAuthReturnParams();
  var initialReturnError = initialReturnParams.state ? initialReturnParams.error : null;
```

Not a blocker.

zeryabkhan91:
Thanks — took the suggestion, and I went and checked the two things you flagged as unknown rather than leaving them as unknowns.

## The mount-time render

Taken as suggested. You're right that it's the same lesson the guard learned and this didn't — I fixed the damage and left the display, which is a partial fix wearing the appearance of a complete one.

## Your two open items — both now answered from code

**The mint does pass CORS.** You were right that my `204` was the redemption, a different origin pair. Checking the mint's pair properly:

`config.allowedOrigins` is built from every region's `host`/`apps_host`/`cdn_host` (`libs/config.js:155-163`) and then **trailing-slash-stripped** at `libs/config.js:199-202`. That last step is what makes it work — `region.host` values carry a trailing slash (`https://us.api.fliplet.com/`) while a browser `Origin` header never does, and `libs/cors.js:36` compares the raw header. Without the normalisation the region-host allowlist would never match a browser request at all.

So a mint from the auth page's origin (an API host) to another region's API host is granted at `cors.js:36`. Credentials aren't needed — the token rides in an explicit `Auth-token` header with no `withCredentials` — and the preflight is fine since `cors.js` sets no `allowedHeaders`.

I nearly reported this as a live CORS bug on the strength of the trailing-slash mismatch before finding the normalisation twenty lines further down.

**`result.user.host` is a full origin, and it is the home region's.** `routes/v1/auth.js:698-710` — when `defaultOrganization.region` differs from the request's region, the login route rebinds both `user` and `region`, so `data.host = config.regions[region].host` reports the **home** region: `https://us.api.fliplet.com/`, trailing slash included. `login.pug:273` already strips it with `.replace(/\/+$/, '')`.

Worth noting the auto-success branch mints against `location.origin`, so it is same-origin and never involved. Only the form branch can go cross-origin, and only after a rebind — which is why this pair was never exercised.

That leaves the region-isolation question (does a *wrong* host fail?) genuinely open, since shared local Redis resolves either way. I've kept it in the PR body as unproven rather than closing it.

## Codacy

Agreed, and your framing is better than mine — it isn't that it's failing, it's that nobody has ever seen what it would say. Six heads, no findings, no title, no summary. It needs a maintainer; I don't have the access.

What I could do is make it not the only opinion in the repo. The Test workflow now also runs **lint**, because a test-only workflow cannot catch a broken `.eslintrc` — and this repo has now shipped two that stopped ESLint starting repo-wide: the invalid `comment` property you saw last round, and an `es2022` env that the lockfile's pinned ESLint 8.2.0 rejects (@Arpanexe caught that one). Neither was visible to CI. Both were mine, and both were things I called clean without running them.

Adding it was instructive. Its first run failed with **13 errors, 12 of which predate this branch** — 6 missing JSDoc `@returns` plus an undeclared `Uint8Array` in `js/build.js`, 4 style errors in `js/interface.js`, 1 in `js/app.js`, all confirmed present on `projects/DEV-1633`. Lint has never run in CI here, so that backlog has simply never been visible.

I scoped the gate to `js/safe-auth-host.js` and `test/` rather than fix them, and left the unscoped run as `lint:all`. Clearing them means editing `js/interface.js` and adding JSDoc across `js/build.js`, which does not belong in a diff that changes the authentication return leg — but I would rather say that plainly than present a narrowed gate as a green one. Worth its own ticket. The scoping does not cost the thing it was added for: an invalid config fails the step whatever files it is pointed at.

The 13th was mine — `no-script-url` on the `javascript:` entry in the hostile-input list, where being a script URL is the point of the assertion. Disabled inline with the reason. `js/safe-auth-host.js` itself came back clean.

## PR description

Updated — the `options.apiUrl`, old-core degradation, no-CI and cached-`navigate.js` claims were all stale. It now also records the two-origin-pair CORS detail above, and the exchange-profile field removal.

